PR #3, opened 2020-06-12
Fix go.sum

Tyilo:
Without this I can't build with `GOFLAGS=-mod=readonly`.


Foxboron:
Frankly. I suspect the issue is because I have done unholyness to my `$GOPATH` while implementing and debugging pkcs7 and misc libraries. Evidently, there are a few attempts in master fixing it :p

PR #9, opened 2020-06-21
Create cmd/sbctl, move main.go there.

### LINKED ISSUE #8: Follow recommended project layout
According to https://github.com/golang-standards/project-layout, commands should go in `cmd/<command-name>/`, and all supporting files in either `pkg/` or `internal/`, depending on if you wish to export them to other Go projects. That would allow the `sbctl` command to be installed with `go install ./...`, which is considerably easier.

I would be interested in making these changes if you agree to them, just need to know where you'd prefer the supporting files to be (pkg or internal).

ericonr:
This allows a user or developer to build and install the command with:

- go build ./cmd/sbctl; and
- go install ./cmd/sbctl

Fixes #8

---

Should I add "Signed off by..."?


Foxboron:
I to use "Signed-off-by" and sign all commits (`git commit -s -S`). But this is more a personal preference

Issue #10, opened 2020-06-23
Sign the fwupmgr bootloader


ericonr:
Hey there! I know the `verify` command checks files that are in the ESP partition, but there are other pieces of software that need to be signed and aren't in the ESP. The example I have is the `fwupdmgr` bootloader, which lives in `/usr/libexec/fwupd/efi/fwupdx64.efi` (on Void, I don't remember where it was on Arch). The issue is that, when trying to check for updates, fwupdmgr fails in the system firmware part on a secure boot protected device, because it can't find the signed version of the bootloader.

```
$ fwupdmgr get-devices
[...]
├─System Firmware:
│     Device ID:           xxxx
│     Current version:     xxxx
│     Minimum Version:     xxxx
│     Vendor:              Dell Inc. (DMI:Dell Inc.)
│     Update Error:        missing signed bootloader for secure boot: /usr/libexec/fwupd/efi/fwupdx64.efi.signed cannot be found
│     GUID:                xxxx
│     Device Flags:        • Internal device
│                          • Requires AC power
│                          • Needs a reboot after installation
│                          • Cryptographic hash verification is available
[...]
```

The part about it failing is probably a bug on their part, which I opened an issue about (https://github.com/fwupd/fwupd/issues/2219), but I don't think it makes sense for them to implement some way of signing the bootloader. Do you think this is something that could be included in `sbctl`? If yes, what would the mechanism for it be? Some system wide configuration file / configuration directory, or hard coded behavior?

Thanks!

Foxboron:
`sbctl sign -s /usr/lib/fwupd/efi/fwupdx64.efi -o /usr/lib/fwupd/efi/fwupdx64.efi.signed`

Should solve all of it. `sbctl` would track `fwupdx64.efi` for changes and write out `fwupdx64.efi.signed` when you sign.

ericonr:
Indeed. Thanks for the quick answer!

Actually, it's there in the README, no idea how I missed it at first.

Foxboron:
I did actually intentionally add it to the README.md for this reason :) It's a nice feature and demonstrates the flexibility.

ericonr:
Ok, I'm fairly sure this has some bug in functionality now...

```
$ cd /usr/libexec/fwupd/efi
$ sudo sbctl -s fwupdx64.efi -o fwupdx64.efi.signed
-> Signing /usr/libexec/fwupd/efi/fwupdx64.efi...
$ sudo sbctl verify
[...]

PR #11, opened 2020-06-24
Update docs

ericonr:


ericonr:
I think this can be left open while decisions are made in #10, that way I can include some new info here, if necessary.

ericonr:
@Foxboron I added a `NOTES` section to explain that all paths are converted. Do you think the phrasing is fine? If so, I think this PR is ready.

Foxboron:
Yas, it's great :) Thanks

PR #15, opened 2020-07-10
Fix SignFile and VerifyESP

ericonr:
SignFile was performing the wrong verification to see if it could skip signing a file, and VerifyESP was printing the wrong error message when the output file couldn't be found.


Foxboron:
Thanks :)

PR #16, opened 2020-07-10
Propagate errors

ericonr:
~~This one includes the commits from #15, it was just easier since I was messing with the code in the same branch :p~~

I tried to do error propagation using mostly the `error` type. We could discuss other actions, such as `verify`, also exiting with error codes.

### DIFF
--- a/sbctl.go
+++ b/sbctl.go
@@ -2,6 +2,7 @@ package sbctl
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"io/ioutil"
 	"log"
@@ -101,11 +102,12 @@ func VerifyESP() {
 	}
 }
 
-func Sign(file, output string, enroll bool) {
+func Sign(file, output string, enroll bool) error {
 	file, err := filepath.Abs(file)
 	if err != nil {
 		log.Fatal(err)
 	}
+
 	if output == "" {
 		output = file
 	} else {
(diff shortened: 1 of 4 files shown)


sbctl.go
```diff
@@ -214,14 +230,25 @@ func CreateBundle(bundle Bundle) {
...
 
-func GenerateAllBundles() {
+func GenerateAllBundles() error {
 	msg.Println("Generating EFI bundles....")
 	bundles := ReadBundleDatabase(BundleDBPath)
+	out := true
 	for _, bundle := range bundles {
-		CreateBundle(*bundle)
+		if !CreateBundle(*bundle) {
+			out = false
```
Foxboron, line 242: You wouldn't know which bundle fails in this case. That might be a bit confusing?

`Failed to write bundle /boot....` would maybe be better?

ericonr: Good idea. Still, bundle generation still has  a lot of errors to propagate. CombineFiles, for example, still calls `log.Fatal`.

ericonr: I printed the error as well, but it might be better to make GenerateBundle return an `error`.

ericonr:
We could add some documentation regarding return error as well, once the error propagation story is decided.

Foxboron:
Yes, I'm all for a well structured man page. Merge this now and open an issue on exit codes and so on?

ericonr:
I'm not completely satisfied with the current state of this one, because there are still quite a few spots where it can abort and exit instead of only propagating errors. I will see if I can try to get it done today.

If you could open an issue with your thoughts on exit codes, that would be great!

sbctl.go
```diff
@@ -214,14 +230,26 @@ func CreateBundle(bundle Bundle) {
...
+			err2.Printf("failed to write bundle %s!", bundle.Output)
+		}
+	}
+
+	if out {
+		return nil
+	} else {
+		msg := "Error generating EFI bundles"
+		err2.Println(msg)
+		return errors.New(msg)
```
Foxboron, line 252: `if !out` then you can throw the `return nil` at the end of the function. Might look more natural?

ericonr: It does, thanks!

ericonr:
Ok, I made some more changes to the error approach. I am not sure if it's the best design, so let me know what you think should change.

I haven't messed around with other `log.Fatal*` functions in the code, but they don't necessarily have to be propagated, since a failure in them should be an application wide failure, like not being able to open the database.

Foxboron:
I'm a little bit unsure what I think about the `PrintGenerateError`. It works for a first iteration, but `PrintGenerateError(fmt.Sprintf(...), logger)` isn't super elegant, but unsure if it would be better with a `func PrintGenerateErrorf(msg string, log logger, args ...string) error`. However it's not a blocker by any means if you are happy :)

ericonr:
I would prefer your version, tbh. Perhaps `func PrintGenerateErrorf(log logger, msg string, args ...string) error`, to keep the string and the args together?

ericonr:
I think these changes are good to go, if you agree with them. I haven't tested the bundle code paths extensively, though.

Foxboron:
did a quick rebase and merged.

PR #21, opened 2020-07-23
main: verify if files exist before removing them.

### LINKED ISSUE #17: Remove-file doesn't error out when removing file that doesn't exist.
Commands like the ones below don't print any error messages. Since we are only saving files with the absolute path, a non absolute path for `remove-file` should either be an error, or it should  get the absolute path before trying to remove the file. If the absolute path doesn't exist we shoud get an error, IMO.

We could omit feedback for the successful case.

```
$ sbctl remove-file file
$ sbctl remove-file /non/existent/file
```

ericonr:
Errors out if trying to remove-file or remove-bundle a file that isn't
in the database. Needs a better log message format.

Fixes #17

### DIFF
--- a/cmd/sbctl/main.go
+++ b/cmd/sbctl/main.go
@@ -107,6 +107,10 @@ func removeFileCmd() *cobra.Command {
 				log.Fatal("Need to specify file")
 			}
 			files := sbctl.ReadFileDatabase(sbctl.DBPath)
+			if _, ok := files[args[0]]; !ok {
+				log.Printf("File %s doesn't exist in database!\n", args[0])
+				os.Exit(1)
+			}
 			delete(files, args[0])
 			sbctl.WriteFileDatabase(sbctl.DBPath, files)
 		},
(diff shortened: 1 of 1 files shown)


cmd/sbctl/main.go
```diff
@@ -243,8 +248,14 @@ func removeBundleCmd() *cobra.Command {
 				log.Fatal("Need to specify file")
 			}
 			bundles := sbctl.ReadBundleDatabase(sbctl.BundleDBPath)
-			delete(bundles, args[0])
-			sbctl.WriteBundleDatabase(sbctl.BundleDBPath, bundles)
+
+			if _, ok := bundles[args[0]]; ok {
```
Foxboron, line 252: Lets invert to `!ok` and drop the `else`-clause :smile:

cmd/sbctl/main.go
```diff
@@ -107,8 +107,13 @@ func removeFileCmd() *cobra.Command {
 				log.Fatal("Need to specify file")
 			}
 			files := sbctl.ReadFileDatabase(sbctl.DBPath)
-			delete(files, args[0])
-			sbctl.WriteFileDatabase(sbctl.DBPath, files)
+			if _, ok := files[args[0]]; ok {
```
Foxboron, line 110: Lets invert to `!ok` and drop the `else`-clause :smile:

ericonr:
I thought about doing that, somehow completely forgot :p 

Thanks!

Foxboron:
I generally like writing if-statements as "I don't want this", this avoid a lot of the nested clauses and `if`/`else`. Shorter code as a bonus :smile:

PR #30, opened 2020-10-27
add travis ci

syscll:
This PR adds the following:
- Travis CI config file
- Bumps Go to v1.15

It will require @Foxboron to enable the repo at https://travis-ci.org.

I would also recommend enabling the following Travis settings (if not already):
- [ ] Build pushed branches
- [ ] Build pushed pull requests
- [ ] Auto cancel pull request builds

And the following GitHub settings (if not already), Branches -> master:
- [ ] Require pull request reviews before merging
- [ ] Dismiss stale pull request approvals when new commits are pushed
- [ ] Require status checks to pass before merging
- [ ] Require branches to be up to date before merging


Foxboron:
I'll sit down and consider travis. The problem is that for integration tests we need to either mock some secure boot files to test functionality or only have local CI to spin up empheral VMs to check this. We also don't have tests anywhere yet so low priority.

It might be better to add staticcheck and `go vet` though?

syscll:
I think this is all achievable with Travis, we could even use Docker to spin up ephemeral containers.

My next PR was going to be the addition of some unit tests, but I'm totally onboard with having a discussion around CI first.

Foxboron:
Please do submit the PRs for some unittesting of the code :) Be vary that I have plans to rip out `sbsigntools` with https://github.com/Foxboron/goefi down the line, so any interactions with those Execs can preferably be omitted.

syscll:
Cool, I will close this for now and keep the branch around in case it's useful in the future.

Will open an issue for a discussion on CI.

Foxboron:
I still think reworking the CI to have `go vet` and `staticcheck` for some smoke testing of the added code is a nice improvement though. So feel free to reopen the PR or submit a new one :)

syscll:
Have added the `staticcheck` cmd to `go.mod` (so we can track versions) and also trigger it in Travis.

Foxboron:
Actually, can we do a 180 on travis and try use github actions instead? One less system and better integrations. I wonder if there are containers with staticcheck we can use instead of the awk in `tools.go` as it would imply a staticcheck vendor in the project when it's not needed.

Thanks for doing this :)

syscll:
Definitely up for trying GitHub actions! It's been on my list for a while.

RE `tools.go`: we could definitely use Docker for `staticcheck`, but the reason for the entry in `go.mod` is to track the specific versions of tools we use, e.g. `staticcheck 2020.1.6`, not just application dependencies.

I will close this PR for now and have a mess around with actions to see how easy it is.

Foxboron:
Lovely. Sorry for the back and forth!

PR #31, opened 2020-10-28
main: help subcommand should work without root.

ericonr:
I'm not sure I need to check the array length, because I don't think it enters the function if there isn't any command at all. But I didn't want to risk it.

Could be changed to `strings.Fields(c.CommandPath())[1]` directly.

### DIFF
--- a/cmd/sbctl/main.go
+++ b/cmd/sbctl/main.go
@@ -200,6 +200,16 @@ func bundleCmd() *cobra.Command {
 			if err != nil {
 				log.Fatal(err)
 			}
+			// Fail early if user wants to save bundle but doesn't have permissions
+			var bundles sbctl.Bundles
+			if save {
+				// "err" needs to have been declared before this, otherwise it's necessary
+				// to use ":=", which shadows the "bundles" variable
+				bundles, err = sbctl.ReadBundleDatabase(sbctl.BundleDBPath)
+				if err != nil {
+					log.Fatalln(err)
+				}
+			}
 			bundle.Output = output
 			bundle.IntelMicrocode = intelucode
 			bundle.AMDMicrocode = amducode
@@ -212,7 +222,6 @@ func bundleCmd() *cobra.Command {
 			bundle.ESP = espPath
 			sbctl.CreateBundle(*bundle)
 			if save {
-				bundles := sbctl.ReadBundleDatabase(sbctl.BundleDBPath)
 				bundles[bundle.Output] = bundle
 				sbctl.WriteBundleDatabase(sbctl.BundleDBPath, bundles)
 				sbctl.FormatBundle(bundle.Output, bundle)
@@ -266,7 +275,10 @@ func removeBundleCmd() *cobra.Command {
 			if len(args) < 1 {
 				log.Fatal("Need to specify file")
 			}
-			bundles := sbctl.ReadBundleDatabase(sbctl.BundleDBPath)
+			bundles, err := sbctl.ReadBundleDatabase(sbctl.BundleDBPath)
+			if err != nil {
+				log.Fatalln(err)
+			}
 
 			if _, ok := bundles[args[0]]; !ok {
 				log.Printf("Bundle %s doesn't exist in database!\n", args[0])
--- a/cmd/sbctl/main.go
+++ b/cmd/sbctl/main.go
@@ -1,11 +1,9 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/foxboron/sbctl"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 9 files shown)


ericonr:
Could also try to use another function that takes the command directly, which probably exists.

cmd/sbctl/main.go
```diff
@@ -307,10 +307,12 @@ func completionFishCmd() *cobra.Command {
 
 func main() {
 	rootCmd.PersistentPreRun = func(c *cobra.Command, args []string) {
+		fields := strings.Fields(c.CommandPath())
 		if strings.Contains(c.CommandPath(), "completion zsh") ||
 			strings.Contains(c.CommandPath(), "completion bash") ||
 			strings.Contains(c.CommandPath(), "completion fish") ||
-			strings.Contains(c.CommandPath(), "__complete") {
+			strings.Contains(c.CommandPath(), "__complete") ||
+			len(fields) > 1 && fields[1] == "help" {
```
ericonr, line 315: I wanted to use `Fields` to protect against the case where a user has `help` in some path passed to the other commands. If that isn't necessary, we should be fine using `Contains` directly.

`cobra` has `c.Name()`, but that gets hairy, since the completion commands are actually subcommands, so it initially returns the shell name.

Foxboron: I'm wondering if we should swap the logic around. Instead of `PersistentPreRun` we can do a `PreRun` on all commands that require root. This can be a fairly simple:

```
for _, cmd := range []interface{createKeysCmd(), enrollKeysCmd()}{
    cmd.Prerun = func(....iforget) {
           // if check
    }
    rootCmd.AddCommand(cmd)
}
```

Add an variable for the commands or something. Wouldn't this be simpler?

ericonr: It has a chance of being much cleaner. I will see if I can implement it.

ericonr:
That said, I was considering removing this check entirely, due to the possibility of adding integration tests. Doing an acess test with the databases and/or keys would allow us to check for adequate permissions without hardcoding the required UID (or getting ugly errors).

For testing (unless it's always done inside a chroot environment, which I don't love), we would require command line options for the `/usr/share/secureboot` folder.

ericonr:
Ping on the idea of moving to some access test instead of checking if uid is 0.

Foxboron:
Yes, a much better idea instead of lazily checking for root.

ericonr:
Ok, I've kinda turned the PR on its head; feel free to point out any issues you have with my changes, since I'm not 100% sure about them either.

I still need to review the changes again, but they look somewhat sane for now.

Foxboron:
I need to pull and test this, but I think this is a better approach then gating the command at the top-level. Allows for more fine-grained permission handling as well! 

Thanks for the work :)

ericonr:
Happy to help :) 

If you don't think the PR will get too bloated, I could start adding stuff for the exit error codes as well.

ericonr:
Ping?

Foxboron (changes requested):

database.go
```diff
@@ -16,24 +18,44 @@ type SigningEntry struct {
...
+				warning.Printf(rootMsg)
+			}
+			return nil, err
 		}
 		file.Close()
+	} else {
+		if os.IsPermission(err) {
+			warning.Printf(rootMsg)
+		}
+		return nil, err
```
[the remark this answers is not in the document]
ericonr: > If err is nil, the first if is False which gives us a return. This code is only handling permission error, but nothing else.

I believe checking explicitly for `err != nil` is best, thanks for pointing this out!

ericonr:
I've dropped my usage of `unix.Access` in `database.go`, since simply checking for errors when trying to read the file is less race-prone and works just as well.

As explained in the commit message, it's still necessary in `keys.go`, since we run `sbsign` as a command.

Foxboron:
The replies are a bit inconsistent:
```
λ sbctl pr-31» ./sbctl verify    
==> WARNING: It might be necessary to run this tool as root
==> WARNING: Couldn't read file database: open /usr/share/secureboot/files.db: permission denied
==> Verifying EFI images in /efi...
  -> WARNING: /efi/EFI/BOOT/BOOTX64.EFI is not signed
  -> WARNING: /efi/EFI/Linux/linux-linux-mitigations.efi is not signed
  -> WARNING: /efi/EFI/Linux/linux-linux.efi is not signed
  -> WARNING: /efi/EFI/arch/fwupdx64.efi is not signed
  -> WARNING: /efi/EFI/systemd/systemd-bootx64.efi is not signed
  -> WARNING: /efi/vmlinuz-linux is not signed
  -> WARNING: /efi/vmlinuz-linux-lts is not signed
λ sbctl pr-31» ./sbctl list-files
==> WARNING: It might be necessary to run this tool as root
  -> ERROR: Couldn't open database: /usr/share/secureboot/files.db
λ sbctl pr-31» ./sbctl list-bundles
2021/01/05 21:07:47 open /usr/share/secureboot/bundles.db: permission denied
```

Part of the problem here is that `ReadBundleDatabase` doesn't implement the access checking. I think it's better at this point to just throw a `ReadFile` function into `util.go` that does this checking. It would allow us to better seek the correct permissions in the future.

ericonr:
Oh, well spotted. I will try to organize it better.

ericonr:
Ok, the PR has increased a bit. I haven't checked extensively if the UI got more consistent, but at least the code should have. It's quite a few changes, but I believe you should be able to follow them logically by looking at the commit history; I tried to explain each change as well as possible.

PR #32, opened 2020-11-02
Add staticcheck linter

syscll:
Based on the comments in #30, this PR adds the `staticcheck` linter and fixes errors found during analysis.


Foxboron:
Can you squash the commits?

syscll:
I think it's easier to squash and merge?

syscll:
Settings -> Merge Button -> Allow squash merging

Foxboron:
I don't merge through github as It does faux signatures on the commits. I can manually pull and squash if it's easier :) Ill be doing it for the merge regardless.

syscll:
Ah, no worries. I don't have access to my laptop atm, so if you wouldn't mind then I'd appreciate it!

README.md
```diff
@@ -1,5 +1,4 @@
-sbctl - Secure Boot Manager
-===========================
+# sbctl - Secure Boot Manager [![Build Status](https://github.com/Foxboron/sbctl/workflows/build/badge.svg)](https://github.com/Foxboron/sbctl/actions)
```
ericonr, line 1: I think it's missing a new line here, isn't it?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/foxboron/sbctl
 
-go 1.14
+go 1.15
```
ericonr, line 3: I'm not sure what the effect of increasing this is...

Makefile
```diff
@@ -47,3 +47,17 @@ push-aur:
...
 	rm -f sbctl
+
+.PHONY: deps
+deps:
+	go get honnef.co/go/tools/cmd/staticcheck@2020.1.6
+
+.PHONY: lint
+lint: deps
+	go vet ./...
+	test -z $(shell go fmt ./...)
```
ericonr, line 58: You might want to quote this, I'm not sure how it behaves in make.

Makefile
```diff
@@ -47,3 +47,17 @@ push-aur:
 clean:
 	rm -f $(MANS)
 	rm -f sbctl
+
+.PHONY: deps
+deps:
+	go get honnef.co/go/tools/cmd/staticcheck@2020.1.6
+
+.PHONY: lint
+lint: deps
```
ericonr, line 56: I would prefer to be able to obtain `staticcheck` on my own; perhaps CI should do `make deps lint` for the dependency, but a user can obtain in in other ways.

README.md
syscll: You would prefer the build status on a new line?

go.mod
syscll: I agree, will revert.

Makefile
syscll: I'm not actually sure this is the best method of doing this, so I might revert.

Makefile
syscll: I would have to disagree here as that I think linting tools should be versioned just like dependencies. That being said, I can remove the call to `deps` when running `lint`, that way you can still use your own version.

Makefile
syscll: `go fmt` annoyingly doesn't return error codes.

syscll:
PTAL

Makefile
ericonr: I don't disagree with the versioning, definitely. That sounds good to me.

README.md
ericonr: I think so, @Foxboron might disagree with me.

syscll:
Replaced by #35

PR #34, opened 2020-11-07
[WIP] keys: use goefi instead of calling sbsiglist.

ericonr:
Calling `sbsiglist` manually resulted in exactly the same file as generated here (it also resulted in my discovery of a segfault in `sbsiglist` :P ). However, I think it would be nice to have more testing in goefi for any regressions and corner cases; furthermore, I think the API could be improved. Perhaps `WriteSignatureList` could take an `io.Writer` or similar?

This PR also doesn't do any of the necessary error checking.

For that matter, I would also change the signature of `KeyToSiglist()`.


Foxboron:
The goefi API needs a lot more work. I'm hesitant using it for sbctl until there is better regression testing. Apart from that, cleary UUID should take bytes, and some better input types into the API. 

Nice to see how integrating it into the codebase would look like though. Makes it a bit easier to spot pain points.

Foxboron:
This is now solved through https://github.com/Foxboron/sbctl/pull/74 :)

PR #36, opened 2020-11-08
Fix lint step in CI.

ericonr:
Set GOBIN during `make deps` so staticcheck is installed directly in a
directory already in PATH.


ericonr:
Merging it in seems to have worked!

Foxboron:
It werks! Lovely :)

PR #38, opened 2020-11-10
Document the generate option of sign-all

physkets:


Foxboron:
Thanks for the patch!

PR #54, opened 2021-04-14
Use x/sys/unix for ioctl instead of rolling our own.

ericonr:
Keep FS_* constants which aren't listed in the package, and leave
SetAttr there, even though it's currently not used.

### DIFF
--- a/chattr.go
+++ b/chattr.go
@@ -1,35 +1,8 @@
-// -*- Mode: Go; indent-tabs-mode: t -*-
-
-/*
- * Copyright (C) 2016 Canonical Ltd
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License version 3 as
- * published by the Free Software Foundation.
- *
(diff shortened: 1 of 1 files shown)


ericonr:
Hm, I'm not 100% about `SetAttr`, actually. It might work just fine on little endian system, but technically speaking it's wrong. The man page is clear about the syscall taking a `int *` and not a `long *` (and C `int` is Go `int32`)

Foxboron:
Ah, yes. Much better. Should probably learn to use the ioctl stuff myself instead of stealing wrappers from canonical :p I'll pull it and test it a bit :)

ericonr:
https://github.com/golang/go/issues/45585 <-- caveat about using these functions; shouldn't be an issue for what platforms sbctl aims to support, but I think a comment is warranted, or I could use the (ugly) workaround I listed in the issue.

The only platform I can think of that'd be affected by this is big endian aarch64.

ericonr:
I've added a comment.

PR #58, opened 2021-05-04
Use `/etc` database path with `/usr` as fallback

### LINKED ISSUE #57: Shouln't `/etc` be a place to store configuration instead of `/usr`
> /usr is the second major section of the filesystem. /usr is shareable, read-only data. That means that /usr should be shareable between various FHS-compliant hosts and must not be written to. Any information that is host-specific or varies with time is stored elsewhere.

[From Linux Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#PURPOSE18)

`/usr` might be useful if you have machines with same secure boot keys but I think for a host specific configuration `/etc` should be used.

igo95862:
`/usr` fallback is for backward compatability.

Closes #57

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -23,8 +24,36 @@ import (
 
 var RSAKeySize = 4096
 
+func getDatabasePath() string {
+	etcDatabasePath := "/etc/secureboot"
+	usrDataBasePath := "/usr/share/secureboot"
+
+	_, err := os.Stat(etcDatabasePath)
+	if err == nil {
+		// Etc database path exists. Use it.
+		return etcDatabasePath
+	} else if errors.Is(err, os.ErrNotExist) {
+		// Ignore not exists error
+	} else {
+		log.Fatalf("Unknown error happended while quering /etc database path: %v", err)
+	}
+
+	_, err = os.Stat(usrDataBasePath)
+	if err == nil {
+		// Usr database path exists. Use it.
+		return usrDataBasePath
+	} else if errors.Is(err, os.ErrNotExist) {
+		// Ignore not exists error
+	} else {
+		log.Fatalf("Unknown error happended while quering /usr database path: %v", err)
+	}
+
+	// Neither /usr nor /etc exitsts. Default to /etc
+	return etcDatabasePath
+}
+
 var (
-	DatabasePath = "/usr/share/secureboot/"
+	DatabasePath = getDatabasePath()
 	KeysPath     = filepath.Join(DatabasePath, "keys")
 	PKKey        = filepath.Join(KeysPath, "PK", "PK.key")
 	PKCert       = filepath.Join(KeysPath, "PK", "PK.pem")


igo95862:
I have not tested it yet. Plan to do so on my new build. (which I was planning to do secure boot on)

Those go packages take their sweet time to download.

Foxboron:
Keep in mind that 0.3 has broken key enrollment. Fixed with https://github.com/Foxboron/sbctl/commit/9060461b35c5acf33d71a968de5e1710eae23b53

igo95862:
```
==> Creating secure boot keys...
  -> Created UUID f103aa18-9aa0-4d7d-9dce-a0d890406729...
==> Create EFI signature list /etc/secureboot/keys/PK/PK.der.esl...
==> Signing /etc/secureboot/keys/PK/PK.der.esl with /etc/secureboot/keys/PK/PK.key...
==> Create EFI signature list /etc/secureboot/keys/KEK/KEK.der.esl...
==> Signing /etc/secureboot/keys/KEK/KEK.der.esl with /etc/secureboot/keys/PK/PK.key...
==> Create EFI signature list /etc/secureboot/keys/db/db.der.esl...
==> Signing /etc/secureboot/keys/db/db.der.esl with /etc/secureboot/keys/KEK/KEK.key...
```

Seems to work. (at least for generating keys)

```
/etc/secureboot
├── GUID
└── keys
   ├── db
   │  ├── db.auth
   │  ├── db.der
   │  ├── db.der.esl
   │  ├── db.key
   │  └── db.pem
   ├── KEK
   │  ├── KEK.auth
   │  ├── KEK.der
   │  ├── KEK.der.esl
   │  ├── KEK.key
   │  └── KEK.pem
   └── PK
      ├── PK.auth
      ├── PK.der
      ├── PK.der.esl
      ├── PK.key
      └── PK.pem
```

keys.go
```diff
@@ -23,8 +24,36 @@ import (
 
 var RSAKeySize = 4096
 
+func getDatabasePath() string {
```
Foxboron, line 27: This doesn't completely work. We use this function before we ever create all the directories. We should instead only return `"/usr/share/secureboot"` if it exists, and in all other cases return `"/etc/secureboot"`

Foxboron (changes requested):

keys.go
igo95862: It works. I tested it on my machine without any directory existing.

Right now it works in this order:
1. `/etc/secureboot` if it exists
2. `/usr/share/secureboot` if it exists
3. `/etc/secureboot` if neither exists

> We should instead only return "/usr/share/secureboot" if it exists, and in all other cases return "/etc/secureboot"

So if both locations exists you want it `/usr` to have higher priority over `/etc`

Foxboron:
I have been thinking a bit and I think I'll close this in favor of actually implementing a config and proper directory usage in sbctl. I don't see the need for this migration to /etc and another one towards a 1.0 release.

I have been mocking up the config parser for `sbctl` and contemplating proper usage of `/etc` and `/var/lib`. It also makes a possible `sbctl migrate` command easier to implement as we are not actively supporting two directories.

This is mostly fueled by my recent progress on go-uefi which I aim to use in sbctl the next upcomming month of two :)

Appreciate the effort on your end though!

igo95862:
> I have been mocking up the config parser

Are you planning on using your own configuration file format?

Foxboron:
No, using the ini format with https://github.com/go-ini/ini. But I wanted a specific behaviour in terms of fallbacks with the values,

PR #61, opened 2021-05-04
Use arrays of arguments instead of spliting by white space 

### LINKED ISSUE #60: Is spliting arguments by white space some legacy code or does it serve a purpose?
```
args := fmt.Sprintf("--cert %s %s", cert, file)
cmd := exec.Command("sbverify", strings.Split(args, " ")...)
```
Is same as:
```
cmd := exec.Command("sbverify", "--cert", cert, file)
```
But with support for white space in directories and file names.

igo95862:
Second commit changes the place of `.splash` argument to be last instead of in the middle but I dont' believe that will change the result. I through the man page of `objcopy` and there is no mention of order at `--add-section` or `--change-section-vma`.

Closes #60

### DIFF
--- a/bundles.go
+++ b/bundles.go
@@ -98,6 +98,7 @@ func GenerateBundle(bundle *Bundle) bool {
 	args = append(args, bundle.EFIStub, bundle.Output)
 	cmd := exec.Command("objcopy", args...)
 	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		if exitError, ok := err.(*exec.ExitError); ok {
 			return exitError.ExitCode() == 0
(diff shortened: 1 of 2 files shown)


Foxboron (commented):
Could you checksum the UEFI executable with and without this patch? They should be the same if the line doesn't affect the result :)

igo95862:
> Could you checksum the UEFI executable with and without this patch? They should be the same if the line doesn't affect the result :)

I checked and the sha256sum stayed the same.

However, I discovered that I had a trailing whitespace in one of the options. I fixed it and forced pushed.

BTW, I noticed there is a redirect of objcopy stdout but not stderr.

Foxboron:
I'd love to rewrite that to some pure Go stuff. But a bit unsure what `objcopy` really does. Feel free to add stderr as well! Thanks :)

igo95862:
Done

PR #68, opened 2021-05-18
Update man entry for default cmdline

WhyNotHugo:
The default cmdline is `/etc/kernel/cmdline`, but the man page said `/proc/cmdline`.


WhyNotHugo:
Ping?

PR #72, opened 2021-05-30
Improved GetEsp function.

### LINKED ISSUE #44: Improve GetESP()
`GetESP()` in `sbctl.go` assumes that there is only one partition with the ESP type AND that it is mounted. Both assumptions can fail when:
* you have multiple disks, each can have its own ESP (and UEFI can boot from any from its boot menu)
* by default systemd sets an automount for the ESP, so it's not even mounted by default - the `lsblk -o "PARTTYPE,MOUNTPOINT"` will print the UUID without any mount points.

This affects for ex. the `verify` command and probably others too.

the approach in systemd (bootctl) is to try[1] `"/efi"`, `"/boot"` and `"/boot/efi"`, and can be overridden with a `SYSTEMD_ESP_PATH` env var. And then it tries to also access[2] that path before deciding it's really an ESP.

[1] https://github.com/systemd/systemd/blob/c335b7c3f8164677316936c7a22e6fca85be28af/src/shared/bootspec.c#L1154
[2] https://github.com/systemd/systemd/blob/c335b7c3f8164677316936c7a22e6fca85be28af/src/shared/bootspec.c#L1080

sbctl could re-implement this same approach or somehow re-use the output of bootctl.

igo95862:
Now checks /efi,/boot and /boot/efi for gpt partition table,
vfat filesystem and ESP partition GUID.

Closes #44

### DIFF
--- a/sbctl.go
+++ b/sbctl.go
@@ -30,6 +30,13 @@ type LsblkRoot struct {
 // Slightly more advanced check
 func GetESP() string {
 
+	for _, env := range []string{"SYSTEMD_ESP_PATH", "ESP_PATH"} {
+		envEspPath, found := os.LookupEnv(env)
+		if found {
+			return envEspPath
+		}
+	}
+
 	out, err := exec.Command(
 		"lsblk",
 		"--json",
--- a/sbctl.go
+++ b/sbctl.go
@@ -2,6 +2,7 @@ package sbctl
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


Foxboron:
Lovely! Can you also add support for specifying `SYSTEMD_ESP_PATH` and `ESP_PATH` as environment variables? It should override whatever `lsblk` reports.

It would be later used in the config file I'm planning :)

sbctl.go
```diff
@@ -15,28 +16,69 @@ import (
...
-	}
-	out, err := exec.Command("lsblk", "-o", "PARTTYPE,MOUNTPOINT").Output()
+
+	out, err := exec.Command(
+		"lsblk",
+		"--json",
+		"--output", "PARTTYPE,MOUNTPOINT,PTTYPE,FSTYPE").Output()
 	if err != nil {
-		log.Fatal(err)
+		log.Panic(err)
```
Foxboron, line 38: I'll rebase my cli branch on this and add a proper `(string, error)` type.

igo95862:
> Lovely! Can you also add support for specifying `SYSTEMD_ESP_PATH` and `ESP_PATH` as environment variables? It should override whatever `lsblk` reports.

Done. However, there are no checks. It simply returns the variable if it is set.

sbctl.go
```diff
@@ -15,28 +16,78 @@ import (
...
+type LsblkRoot struct {
+	Blockdevices []LsblkEntry `json:"blockdevices"`
+}
+
+// Slightly more advanced check
 func GetESP() string {
-	if _, err := os.Stat("/efi"); !os.IsNotExist(err) {
-		return "/efi"
+
+	envEspPath, found := os.LookupEnv("SYSTEMD_ESP_PATH")
```
Foxboron, line 33: `for _, env := range []string{"SYSTEMD_ESP_PATH", "ESP_PATH"} {`

Less code duplication for the sake of a for loop.

Foxboron (changes requested):

Foxboron:
`systemd` only checks that it's an absolute path and that it exists. Currently `sbctl` can't report any errors like this properly so leaving it up to the user is fine in my book. We can be a bit more friendly later.

Foxboron:
Ah, missing documentation! Do you want to add a section to the manpage or should I just do it before the release?

igo95862:
Can you do it? Thanks.

Foxboron:
Sure, no problem :) Merged locally and did a quick fixup to the default arguments in the bundle command. I'll write the manpage and do a release later. 

Thanks for the work!

PR #77, opened 2021-06-19
logging: use simple unicode symbols

hexchain:
Use simple unicode symbols in logging, instead of (possibly) emojis.

These two symbols should exist in most monospace fonts, so they can be shown with correct colors, without falling back to emojis, in most cases. They are also what systemd uses in their yes/no style output.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -8,8 +8,8 @@ import (
 )
 
 var (
-	OkSym    = "✔"
-	NotOkSym = "✘"
+	OkSym    = "✓"
+	NotOkSym = "✗"
 	WarnSym  = "‼"
 	UnkwnSym = "⁇"
 )


Foxboron:
Awh, but I really liked the look of the fatter ones :/ I'll pull it and take a look. How uncommon are they?

hexchain:
Within the following fonts in my system:
- DejaVu Sans Mono
- Hack
- Source Code Pro
- Roboto Mono
- JetBrains Mono

Only DejaVu Sans Mono seems to have that check mark.

Another thing is that the check mark emoji is usually black (in Noto Color Emoji and Twemoji), which basically disappears on a dark background terminal...

WhyNotHugo:
I'm using `Twemoji` on my system, and the current checkmark is almost-invisible:

2021-06-27T14:42:58,165807798+02:00

It's not a big deal, IMHO, but since there's a discussion on it, that's my ¢2.

Foxboron:
@WhyNotHugo how does it look with this patch?

WhyNotHugo:
2021-06-27T19:26:35,411878429+02:00

WhyNotHugo:
It might not make sense, but I'd add a checkmark to the second item.

The _missing_ checkmark on that line always stands out as something being wrong to me, even though that's not the case.

Issue #78, opened 2021-06-27
`GetESP` returns an empty string


WhyNotHugo:
I'm using [systemd's Discoverable Partitions Specification](https://systemd.io/DISCOVERABLE_PARTITIONS/).

That is, my `efi` partition is not in my `fstab`:

```console
$ cat /etc/fstab
# Static information about the filesystems.
# See fstab(5) for details.

# See https://github.com/systemd/systemd/issues/20019
UUID=e68d67d9-c05c-45c3-93f9-fc499f3e2e64	none      	swap      	defaults	0 0
```

It is auto-mounted by `systemd` based on its partition type:

```console
$ mount | grep efi
efivarfs on /sys/firmware/efi/efivars type efivarfs (rw,nosuid,nodev,noexec,relatime)
systemd-1 on /efi type autofs (rw,relatime,fd=51,pgrp=1,timeout=120,minproto=5,maxproto=5,direct,pipe_ino=13874)
/dev/nvme0n1p1 on /efi type vfat (rw,relatime,fmask=0077,dmask=0077,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro)
```

However, the partition is auto-mounted when it's accessed, so before reading it or accessing it, it's not mounted, and therefore, won't show up in `lsblk`:

```console
$ lsblk --json --output PARTTYPE,MOUNTPOINT,PTTYPE,FSTYPE
{
   "blockdevices": [
      {
         "parttype": null,
         "mountpoint": "/",
         "pttype": null,
         "fstype": "ext4"
      },{
         "parttype": null,
         "mountpoint": "[SWAP]",
         "pttype": null,
         "fstype": "swap"
      },{
         "parttype": null,
         "mountpoint": null,
         "pttype": "gpt",
         "fstype": null
      },{
         "parttype": "c12a7328-f81f-11d2-ba4b-00a0c93ec93b",
         "mountpoint": null,
         "pttype": "gpt",
         "fstype": "vfat"
      },{
         "parttype": "4f68bce3-e8cd-4db1-96e7-fbcaf984b709",
         "mountpoint": null,
         "pttype": "gpt",
         "fstype": "crypto_LUKS"
      },{
         "parttype": "0657fd6d-a4ab-43c4-84e5-0933c84b4f4f",
         "mountpoint": null,
         "pttype": "gpt",
         "fstype": "crypto_LUKS"
      }
   ]
}
```

This results in `GetESP` returning an empty string.

[...]

PR #79, opened 2021-06-27
Use /efi as a fallback if the ESP is not found

WhyNotHugo:
The current fallback is an empty string, which is never correct.

This will be correct in _some_ scenarios, but not perfect.

See #78


WhyNotHugo:
I'll do option 2 from #78, and return an error here if nothing matches, which seems safer.

Foxboron:
Thanks!

PR #80, opened 2021-06-27
Add a new "backup" command

WhyNotHugo:
This new command creates a backup of the currently booted EFI image.

This is ideal to run after a successful boot, and serves as a rescue mechanism: each time the system boots successfully, the entry used to boot is backed up, so if you break it before the next reboot, you have a backup/rescue one.

There's a bundled `systemd` service included to automate this.

---

If this looks good, I've two minor improvements to add here:

- If the source file changes _after_ the system booted, skip backing up.
- If we're currently booted of a backup entry, don't create a backup of the backup.

### DIFF
--- a/cmd/sbctl/backup.go
+++ b/cmd/sbctl/backup.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/foxboron/go-uefi/efi"
+	"github.com/foxboron/sbctl"
+	"github.com/foxboron/sbctl/logging"
+	"github.com/spf13/cobra"
+)
+
+var backupCmd = &cobra.Command{
+	Use:   "backup",
+	Short: "Back up currently booted entry.",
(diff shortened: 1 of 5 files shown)


WhyNotHugo:
As discussed, closing as out-of-scope.

All the useful bit have been upstreamed, but `sb-backup` itself is very specific to my own setup.

Foxboron:
A summary of the IRC discussion:

Essentially this problem is more complicated then just backing up the boot entry and call it a day. We need to figure out kernel differences and restore kernel modules upon boot to not mess up anything. This makes it a lot more painfull to properly implement without versioned kernels.

https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/ also details some ways to recover boot with `sd-boot` but I haven't tried it.

WhyNotHugo:
The `Automatic Boot Assessment` spec seems to be designed for systems where upgrades are unattended, **and** older kernels are kept around after upgrades.

My main use-case for `sb-backup` is to have a backup entry to boot if I, for example, mess up my `cmdline` while experimenting.

PR #81, opened 2021-06-28
Attempt to trigger automounts to find the ESP

### LINKED ISSUE #78: `GetESP` returns an empty string

WhyNotHugo:
Attempt to trigger automounts to find the ESP

The ESP may be an automount partition, so try touching a file in each
candidate location so as to trigger an automounts.

This is the same way systemd attempts to find it:
https://github.com/systemd/systemd/blob/f565b86/src/shared/bootspec.c#L1014-L1018

I've also changed the function to return an error if no ESP is found.
The previous behaviour (an empty string) just results in a crash later
on.

When no ESP is found, the `bundle` command will have no default for the
`esp` flag. Passing an empty string to it as a default results in no
value being show in the output of `--help`.

This seemed like the most reasonable compromise instead of panicking.

Fixes #78

### DIFF
--- a/bundles.go
+++ b/bundles.go
@@ -74,15 +74,18 @@ func GetEfistub() string {
 	return ""
 }
 
-func NewBundle() *Bundle {
-	esp := GetESP()
+func NewBundle() (bundle *Bundle, err error) {
+	esp, err := GetESP()
+	if err != nil {
+		return
+	}
 
 	stub := GetEfistub()
 	if stub == "" {
 		panic("No EFISTUB file found. Please install systemd-boot or gummiboot!")
 	}
 
-	return &Bundle{
+	bundle = &Bundle{
 		Output:         "",
 		IntelMicrocode: "",
 		AMDMicrocode:   "",
@@ -94,6 +97,8 @@ func NewBundle() *Bundle {
 		EFIStub:        stub,
 		ESP:            esp,
 	}
+
+	return
 }
 
 func GenerateBundle(bundle *Bundle) (bool, error) {
--- a/cmd/sbctl/bundle.go
+++ b/cmd/sbctl/bundle.go
@@ -40,7 +40,10 @@ var bundleCmd = &cobra.Command{
 				os.Exit(1)
 			}
 		}
-		bundle := sbctl.NewBundle()
+		bundle, err := sbctl.NewBundle()
+		if err != nil {
+			return err
+		}
 		output, err := filepath.Abs(args[0])
 		if err != nil {
 			return err
@@ -81,7 +84,11 @@ var bundleCmd = &cobra.Command{
 }
 
 func bundleCmdFlags(cmd *cobra.Command) {
-	esp := sbctl.GetESP()
+	esp, err := sbctl.GetESP()
+	if err != nil {
+		logging.Warn("Failed to find ESP: %s", err)
+	}
+
 	f := cmd.Flags()
 	f.StringVarP(&amducode, "amducode", "a", "", "AMD microcode location")
 	f.StringVarP(&intelucode, "intelucode", "i", "", "Intel microcode location")
--- a/sbctl.go
+++ b/sbctl.go
@@ -24,14 +24,27 @@ type LsblkRoot struct {
 	Blockdevices []LsblkEntry `json:"blockdevices"`
 }
 
+var espLocations = []string{
+	"/boot",
+	"/boot/efi",
+	"/efi",
+}
+var ErrNoESP = errors.New("failed to find EFI system partition")
+
 // Slightly more advanced check
-func GetESP() string {
+func GetESP() (string, error) {
 
 	for _, env := range []string{"SYSTEMD_ESP_PATH", "ESP_PATH"} {
 		envEspPath, found := os.LookupEnv(env)
 		if found {
-			return envEspPath
+			return envEspPath, nil
 		}
+	}
+
+	for _, location := range espLocations {
+		// "Touch" a file inside all candiadate locations to trigger an
+		// automount if there's an automount partition.
+		os.Stat(fmt.Sprintf("%s/does-not-exist", location))
 	}
 
 	out, err := exec.Command(
@@ -80,10 +93,10 @@ func GetESP() string {
 			continue
 		}
 
-		return entryToCheck.Mountpoint
+		return entryToCheck.Mountpoint, nil
 	}
 
-	return ""
+	return "", ErrNoESP
 }
 
 func Sign(file, output string, enroll bool) error {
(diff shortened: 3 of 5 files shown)


bundles.go
```diff
@@ -75,7 +75,10 @@ func GetEfistub() string {
 }
 
 func NewBundle() *Bundle {
-	esp := GetESP()
+	esp, error := GetESP()
+	if error != nil {
+		panic(error)
```
Foxboron, line 80: No panics please :)

sbctl.go
```diff
@@ -80,10 +92,10 @@ func GetESP() string {
 			continue
 		}
 
-		return entryToCheck.Mountpoint
+		return entryToCheck.Mountpoint, nil
 	}
 
-	return ""
+	return "", fmt.Errorf("failed to locate ESP. Is it mounted?")
```
Foxboron, line 98: `var ErrNoESP = errors.new("missing EFI system partition")` and `return "", ErrNoESP` then we can give guidance in the top level by matching on the error in `cmd/sbctl/main.go`.

sbctl.go
```diff
@@ -24,16 +24,28 @@ type LsblkRoot struct {
...
 
 	for _, env := range []string{"SYSTEMD_ESP_PATH", "ESP_PATH"} {
 		envEspPath, found := os.LookupEnv(env)
 		if found {
-			return envEspPath
+			return envEspPath, nil
 		}
 	}
 
+	for _, location :=range espLocations {
```
Foxboron, line 43:
```suggestion
	for _, location := range espLocations {
```
Missing space

Foxboron (changes requested):

bundles.go
WhyNotHugo: There's another `panic` two lines down.

Foxboron: Ohno, I thought I had removed all of them! My misake :) sorry!

cmd/sbctl/bundle.go
```diff
@@ -80,8 +83,13 @@ var bundleCmd = &cobra.Command{
 	},
 }
 
-func bundleCmdFlags(cmd *cobra.Command) {
-	esp := sbctl.GetESP()
+func bundleCmdFlags(cmd *cobra.Command) error {
+	esp, err := sbctl.GetESP()
+	if err != nil {
```
WhyNotHugo, line 88: If we fail to auto-detect the `ESP`, then there will be no default value for it when generating bundles.

I think it's the best fallback, since the only other option is to fail completely, and user's won't even be able to see `--help`.

Foxboron: Yep, I agree!

cmd/sbctl/bundle.go
```diff
@@ -80,8 +83,12 @@ var bundleCmd = &cobra.Command{
...
+		logging.Warn("Failed to find ESP: %s", err)
```
Foxboron, line 89: Should be `%w` instead of `%s`

cmd/sbctl/bundle.go
```diff
@@ -80,8 +83,12 @@ var bundleCmd = &cobra.Command{
 	},
 }
 
-func bundleCmdFlags(cmd *cobra.Command) {
-	esp := sbctl.GetESP()
+func bundleCmdFlags(cmd *cobra.Command) error {
```
Foxboron, line 86: Did we need the error signature? It's not really used

cmd/sbctl/bundle.go
WhyNotHugo: Linter doesn't like this.

Issue #82, opened 2021-07-14
private keys shouldn't be world-readable


EvilBit:
All key material, including private keys, are created world-readable ([`0644` here](https://github.com/Foxboron/sbctl/blob/78cdabfa6da24e0fb1da84e03d9f74a1d805faf9/keys.go#L90)) at the moment.

Exposing this secret key material to unprivileged users and processes poses a security vulnerability.

EvilBit:
Oh, and by the way, thanks **a lot** for this great tool - finally a workable secure boot management tool :)

Foxboron:
Yolo, copy-pasted code. Thanks :)

ericonr:
Interesting, my keys are 600... I wonder when the regression happened.

EvilBit:
> Interesting, my keys are 600... I wonder when the regression happened.

I just traversed backwards through `git blame` and the permissions in `SaveKey` were `0644` since the [initial commit](https://github.com/Foxboron/sbctl/blob/0b5d4a46eaf0cd6e2bb36a0a5b2ba46802388b52/keys.go#L71).

But [here](https://github.com/Foxboron/sbctl/commit/1bf2810228431862f301e52ca4121cd0b7805346#diff-a0896c3a091e9111074431503f5eb3a62aa117962efd8f4033f905a2c2df356cL71) the initial creation of the files was removed from `CreateKey` (which set `0600`) and relied now on the erroneous permissions in `SaveKey`.

EvilBit:
Maybe the code should even check for too permissive file permissions and refuse signing in this case. Like OpenSSH [enforces it for private keys](https://github.com/openssh/openssh-portable/blob/2dc328023f60212cd29504fc05d849133ae47355/authfile.c#L106).

ericonr:
I think that's reasonable, but please with better error messages :p

Foxboron:
We can just do an `var ErrInvalidPrivKeyPersm = erros.new("...something")` and match it in `main.go` to give something human readable error back.

https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/main.go#L69-L78

Foxboron:
Fixed with https://github.com/Foxboron/sbctl/commit/ea325ca46fec8b75ac6dd89742d2ea714593d38f

PR #83, opened 2021-07-14
fix: restrict key file permissions

EvilBit:
hotfix for #82

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -87,7 +87,7 @@ func CreateKey(name string) ([]byte, []byte, error) {
 
 func SaveKey(k []byte, file string) error {
 	os.MkdirAll(filepath.Dir(file), os.ModePerm)
-	err := os.WriteFile(file, k, 0644)
+	err := os.WriteFile(file, k, 0400)
 	if err != nil {
 		return err
 	}


keys.go
```diff
@@ -87,7 +87,7 @@ func CreateKey(name string) ([]byte, []byte, error) {
 
 func SaveKey(k []byte, file string) error {
 	os.MkdirAll(filepath.Dir(file), os.ModePerm)
-	err := os.WriteFile(file, k, 0644)
+	err := os.WriteFile(file, k, 0640)
```
WhyNotHugo, line 90: Why not `600`?

Foxboron: I'd say 0600 is better since we do not run `sbctl` as it's own user, thus there is no use to have the root group readable I think?

EvilBit: I actually pondered the same question before deciding on these permissions. My reasoning for `0640` was, that there might be edge cases on some systems which use the `root` group for non-root administrative users.
That being said, `0600` would also have been my first choice without these considerations and I'm happy to change the permissions to be even more restrictive and update the PR.

By the way, what's your opinion on making them even `0400` to gain some safety against accidental deletion? Those keys are somewhat precious after all :)

WhyNotHugo:
If it were up to me, I'd say `600` on creation and `400` after enrollment, but I think that's making things too complicated XD

Either 600 or 400 is fine in my book.

EvilBit:
I changed it to `0400` now. `SaveKeys` is only ever invoked from `InitializeSecureBootKeys`, so read-only should be fine.
Replacing/overwriting keys should be a conscious, confirmed choice.

WhyNotHugo:
No objection here.

PR #97, opened 2021-09-27
Add contrib/kernel-install/91-sbctl.install

eNV25:
Add contrib/kernel-install/91-sbctl.install for [kernel-install(8)](https://man.archlinux.org/man/kernel-install.8).

It will sign the kernel when installed and remove it when removed using kernel-install.

This file should be installed be installed to `/usr/lib/kernel/install.d/91-sbctl.install` by a package, or to `/etc/kernel/install.d/91-sbctl.install` by the user.


Foxboron:
@ericonr Mind reviewing this kernel-install hook :)?

eNV25:
Are you sure it should be added to the Makefile? The pacman hook is not there.

Foxboron:
Hmm, true. Just make an install entry like with the completions and don't add it to the completion task. Still easier then having to remember the install line :)

Foxboron:
If you don't want to respect `PREFIX` you need to use `$(DESTDIR)/usr/share/`. Feel free to fix up the install target in sbctl as well :)

eNV25:
```
.PHONY: completion
completion:
	./sbctl completion bash | install -Dm644 /dev/stdin "$(DESTDIR)$(SHRDIR)/bash-completion/completions/sbctl"
	./sbctl completion zsh | install -Dm644 /dev/stdin "$(DESTDIR)$(SHRDIR)/usr/share/zsh/site-functions/_sbctl"
	./sbctl completion fish | install -Dm644 /dev/stdin "$(DESTDIR)$(SHRDIR)/usr/share/fish/vendor_completions.d/sbctl.fish"
```

eNV25:
Also the PKGBUILD for the AUR and community repo should be revised since the completions are installed in the `install` target. Currently they are installed separately.

PR #104, opened 2021-10-24
Ignore EOF errors in CheckMSDos

### LINKED ISSUE #96: verify does not show all EFI binaries
`sbclt verify` does not show all EFI binaries. Notable missing ones are:

- _/boot/EFI/systemd/systemd-bootx64.efi_
- _/boot/EFI/arch/fwupdx64.efi_

After signing these files, they show up on `sbctl verify` (as they are stored in the database now, I assume).

Note that the ESP is mounted to /boot.

Tested with sbctl version 0.6-1

silkeh:
CheckMSDos attempts to read the first two bytes of a file to determine
if it is an MS-DOS executable. If a file (for whatever reason) is shorter,
attempting to read this will result in an `io.EOF` error.

The error can be safely ignore because such an error indicates that the file
could be read successfully, it just doesn't contain enough bytes.

Possibly resolves #96

### DIFF
--- a/util.go
+++ b/util.go
@@ -105,6 +105,10 @@ func CheckMSDos(path string) (bool, error) {
 	// They contain "MZ" as the two first bytes
 	var header [2]byte
 	if _, err = io.ReadFull(r, header[:]); err != nil {
+		// File is smaller than 2 bytes
+		if errors.Is(err, io.EOF) {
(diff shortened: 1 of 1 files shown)


Foxboron:
Thanks for the patch!

PR #106, opened 2021-11-09
Add convenient aliases for some sub-commands

eNV25:
Add convenient aliases for some sub-commands.

Aliases added:

- `list-bundles`
  - `ls-bundles`
- `list-files`
  - `ls-files`
  - `ls`
- `remove-bundle`
  - `rm-bundle`
- `remove-file`
  - `rm-file`
  - `rm`

If you don't agree with the `ls` and `rm` aliases I can remove them.

### DIFF
--- a/cmd/sbctl/list-bundles.go
+++ b/cmd/sbctl/list-bundles.go
@@ -15,7 +15,10 @@ type JsonBundle struct {
 }
 
 var listBundlesCmd = &cobra.Command{
-	Use:   "list-bundles",
+	Use: "list-bundles",
+	Aliases: []string{
+		"ls-bundles",
+	},
 	Short: "List stored bundles",
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


Foxboron:
Hmm, my first impression is that I like it. I need to think a bit and figure out if this could be confusing in the long run? But I don't see anything obvious at least.

Foxboron:
These aliases should preferably be mentioned in the man page :)

https://github.com/Foxboron/sbctl/blob/master/docs/sbctl.8.txt

ericonr:
I'm slightly against this, because wiki pages and tutorials other folks write about sbctl will end up using multiple different versions of each command, and I think that would be confusing if someone is trying to get a full picture of things.

Idk about you, but I run these rarely enough that length doesn't matter as much...

eNV25:
I thought it would be a good idea because it's similar to `git ls-files` and `git rm`. I added the other aliases for completeness.

PR #126, opened 2022-02-14
read key from private key file

tpeacock19:
Have `sbctl import-keys` read from correct private key files.


Foxboron:
Thanks for the patch!

PR #133, opened 2022-03-24
Remove hardcoded architecture in filename

WhyNotHugo:
Noticed this issue when using `sbctl` to build a bundle on aarch64.

I added these three architectures because they're the ones for which systemd ships a stub; no more, no less.

### DIFF
--- a/bundles.go
+++ b/bundles.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
CI failure seems unrelated: https://github.com/Foxboron/sbctl/runs/5680167905?check_suite_focus=true#step:6:22

Maybe just retry?

PR #151, opened 2022-07-04
Check a few errors that are currently being ignored

WhyNotHugo:


sbctl.go
```diff
@@ -41,9 +41,9 @@ func GetESP() (string, error) {
 	}
 
 	for _, location := range espLocations {
-		// "Touch" a file inside all candiadate locations to trigger an
+		// "Read" a file inside all candiadate locations to trigger an
 		// automount if there's an automount partition.
-		os.Stat(fmt.Sprintf("%s/does-not-exist", location))
+		_, _ = os.Stat(fmt.Sprintf("%s/does-not-exist", location))
```
Foxboron, line 46: Whats the reason to define `_, _ =` in this case? Is this a linter complaint or a personal choice?

util.go
```diff
@@ -172,7 +172,10 @@ func CopyFile(src, dst string) error {
 		return err
 	}
 	defer f.Close()
-	io.Copy(f, source)
+	_, err = io.Copy(f, source)
+	if err != nil {
```
Foxboron, line 176:
```suggestion
	if _, err = io.Copy(f, source); err != nil {
```

sbctl.go
```diff
@@ -55,7 +55,10 @@ func GetESP() (string, error) {
 	}
 
 	var lsblkRoot LsblkRoot
-	json.Unmarshal(out, &lsblkRoot)
+	err = json.Unmarshal(out, &lsblkRoot)
+	if err != nil {
```
Foxboron, line 59:
```suggestion
	if err = json.Unmarshal(out, &lsblkRoot); err != nil {
```

cmd/sbctl/list-files.go
```diff
@@ -55,9 +55,9 @@ func RunList(_ *cobra.Command, args []string) error {
 		return err
 	}
 	if cmdOptions.JsonOutput {
-		JsonOut(files)
+		err = JsonOut(files)
```
Foxboron, line 58: Do a return here?

sbctl.go
WhyNotHugo: Linter complaint: unchecked `err`.

It's almost always a very valid lint, this is a very unusual exception.

cmd/sbctl/list-bundles.go
```diff
@@ -69,9 +69,9 @@ var listBundlesCmd = &cobra.Command{
 			return err
 		}
 		if cmdOptions.JsonOutput {
-			JsonOut(bundles)
+			err = JsonOut(bundles)
```
Foxboron, line 72: Return here?

bundles.go
```diff
@@ -33,7 +33,10 @@ func ReadBundleDatabase(dbpath string) (Bundles, error) {
 		return nil, err
 	}
 	bundles := make(Bundles)
-	json.Unmarshal(f, &bundles)
+	err = json.Unmarshal(f, &bundles)
+	if err != nil {
```
Foxboron, line 37:
```suggestion
	if err = json.Unmarshal(f, &bundles); err != nil {
```

certs/certs.go
```diff
@@ -42,7 +42,10 @@ func GetCerts(oem string) (*signature.SignatureDatabase, error) {
 			continue
 		}
 		buf, _ := content.ReadFile(path)
-		sigdb.Append(signature.CERT_X509_GUID, GUID, buf)
+		err := sigdb.Append(signature.CERT_X509_GUID, GUID, buf)
+		if err != nil {
```
Foxboron, line 46:
```suggestion
		if err := sigdb.Append(signature.CERT_X509_GUID, GUID, buf); err != nil {
```

cmd/sbctl/status.go
```diff
@@ -88,7 +88,10 @@ func RunStatus(cmd *cobra.Command, args []string) error {
 		stat.Vendors = keys
 	}
 	if cmdOptions.JsonOutput {
-		JsonOut(stat)
+		err := JsonOut(stat)
+		if err != nil {
```
Foxboron, line 92:
```suggestion
		if err := JsonOut(stat); err != nil {
```

database.go
```diff
@@ -21,7 +21,10 @@ func ReadFileDatabase(dbpath string) (SigningEntries, error) {
 	}
 
 	files := make(SigningEntries)
-	json.Unmarshal(f, &files)
+	err = json.Unmarshal(f, &files)
+	if err != nil {
```
Foxboron, line 25:
```suggestion
	if err = json.Unmarshal(f, &files); err != nil {
```

Foxboron (changes requested):

Foxboron:
Okay, I didn't bother correcting all the `if err...; err != nil {` stuff in the review :p sorry!

Generally looks okay something we should have been doing for a while :)

sbctl.go
Foxboron: Ahh! Makes sense :) Then it's fine

WhyNotHugo:
> Generally looks okay something we should have been doing for a while :)

I've been using [`golangci-lint`](https://golangci-lint.run/) to sniff all these out quickly:

```sh
> cat =golint 
#!/bin/sh

podman run --rm -v "$(pwd):/src" -w /src \
  -v "${GOPATH:-$HOME/go}:/go" \
  docker.io/golangci/golangci-lint:v1.46 \
  golangci-lint run --color=always
```

gopls is also very helpful

PR #157, opened 2022-08-05
Drop checksum field from signed files db

WhyNotHugo:
This is required before moving files from `/usr` into their proper
locations.

See: https://github.com/Foxboron/sbctl/issues/57

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
       container: voidlinux/voidlinux-musl
       steps:
          # update xbps, if necessary
-         - run: xbps-install -Syu || ( xbps-install -yu xbps && xbps-install -yu )
+         - run: xbps-install -Syu || ( xbps-install -yu xbps && xbps-install -Syu)
          - run: xbps-install -y make go asciidoc gcc git
          - uses: actions/checkout@v1
          - run: git config --global --add safe.directory $(pwd)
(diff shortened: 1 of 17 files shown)


WhyNotHugo:
Still in draft due to:

- One big comment in the middle regarding which feedback is welcome.
- One test is still failing.
- Tests need a locally EFI binary as input. I'm using a system one, but then tests will fail in other setups.

WhyNotHugo:
Regarding this bit https://github.com/Foxboron/sbctl/blob/e742a7fbdd1a92b5a81cb15baba8d289b24830e8/keys.go#L150-L188, I've doubts.

Is stripping the signature and re-signing a valid way to determine "it's already signed"? If the signature mismatches, we can either bail or overwrite. I think exiting non-zero is best?

Foxboron:
>Is stripping the signature and re-signing a valid way to determine "it's already signed"? If the signature mismatches, we can either bail or overwrite. I think exiting non-zero is best?

Can't we use `pkcs7.VerifySignature`? If we can't verify the file then we probably havent signed it. It seems simpler than messing with the implementation details of the weird context struct :)

Foxboron:
I realized I might not even correctly remember the contraints we are working with :/ I really need to get back into hacking on all of this.

WhyNotHugo:
We also need to verify whether the file has changed or not. The file might have a valid signature, but be a _different_ file (e.g.: the previous version of the bootloader).

PR #169, opened 2022-10-18
update manpage to reflect defaults in bundle.go

SebinNyshkim:
`cmd/sbctl/bundle.go` sources different default paths for initramfs and kernel image location than what the manpage states `sbctl bundle` will source.


Foxboron:
Thanks!

PR #173, opened 2022-11-12
Add example workflow with real screenshots

medhefgo:


medhefgo:
@Foxboron Here you go. Feel free change the wording and or re-arrange the whole thing and move it wherever you like and force push to the branch (I suck at prose). Let me know if you want some more screenshots for more stuff.

medhefgo:
Also, I think the status page could use some rework. It is very confusing with setup mode:
```
Setup Mode:     ✗ Enabled
Secure Boot:    ✗ Disabled
```

Imho, the ✗ when in setup mode is really off-putting. Also, in general the two modes are mutually exclusive, so it makes more sense to reflect this by making this an enum like `bootctl` does now.

Foxboron:
They are largely an artifact of the fact that they are two independant variables being parsed. Thus the obvious way is to have them as two different strings in the output.

I do agree it's not great atm :) 

Thanks! I'll review this soon and maybe copy-paste the doc page for this into a wiki page on the project.

Foxboron:
Thanks for doing this :) I'll probably throw this into a wiki-page of some sort as well.

PR #176, opened 2022-11-25
sbctl.8: Provide more precide setup mode instructions

medhefgo:
The previous instructions were completely wrong: To entroll custom keys
the firmware has to be in setup mode. By enrolling our own keys we enter
user mode! Also, setup mode is automatically entered on PK enrollment,
requiring no further user interaction in the firmware.


Foxboron:
Heh, writing is hard :)

Thanks for fixing this!

PR #183, opened 2022-12-19
Fix POSIX sh comparison

swsnr:
In POSIX sh == doesn't really exist; shellcheck warns about this.


swsnr:
Alternatively I could also change the shebang to bash… what would you prefer?

eNV25 (approved):

Foxboron (approved):

Foxboron:
@eNV25 thanks for the review :)

Issue #187, opened 2023-01-09
kernel install hook should do nothing if keys do not exist


ajakk:
Hi! I have `sbctl` installed in various Gentoo systems, not all of which have keys generated, probably most notably containers. In these containers, kernels are built and installed as part of a normal Gentoo upgrade process with various external kernel modules. Of course, containers have no need for their own secureboot keys, but `sbctl`'s kernel install hook tries to sign the kernels anyway, which results in the upgrade process eventually exiting in error:

```
dracut: *** Creating image file '/usr/src/linux-5.15.85-gentoo-dist/arch/x86/boot/initrd' ***
dracut: dracut: using auto-determined compression method 'gzip'
dracut: *** Creating initramfs image file '/usr/src/linux-5.15.85-gentoo-dist/arch/x86/boot/initrd' done ***
 [ ok ]
 * Installing the kernel via installkernel ...
couldn't access /usr/share/secureboot/keys/db/db.pem: no such file or directory
 [ !! ]
 * ERROR: sys-fs/zfs-kmod-2.1.7::gentoo failed (postinst phase):
 *   Installing the kernel failed
 *
 * Call stack:
 *     ebuild.sh, line  136:  Called pkg_postinst
 *   environment, line 3031:  Called dist-kernel_reinstall_initramfs '/usr/src/linux' '5.15.83-gentoo-dist-hardened'
 *   environment, line 1331:  Called dist-kernel_install_kernel '5.15.83-gentoo-dist-hardened' '/usr/src/linux/arch/x86/boot/bzImage' '/usr/src/linux/System.map'
 *   environment, line 1310:  Called die
 * The specific snippet of code:
 *       eend ${?} || die -n "Installing the kernel failed"
 *
 * If you need support, post the output of `emerge --info '=sys-fs/zfs-kmod-2.1.7::gentoo'`,
 * the complete build log and the output of `emerge -pqv '=sys-fs/zfs-kmod-2.1.7::gentoo'`.
 * The complete build log is located at '/var/tmp/portage/sys-fs/zfs-kmod-2.1.7/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sys-fs/zfs-kmod-2.1.7/temp/environment'.
 * Working directory: '/var/tmp/portage/sys-fs/zfs-kmod-2.1.7/empty'
 * S: '/var/tmp/portage/sys-fs/zfs-kmod-2.1.7/work/zfs-kmod-2.1.7'
[...]

PR #188, opened 2023-01-10
91-sbctl.install: don't sign without signing keys

ajakk:
It's expected that signing doesn't work without having previously generated keys, so don't try to sign when keys don't exist.

Closes: https://github.com/Foxboron/sbctl/issues/187
Signed-off-by: John Helmert III <ajak@gentoo.org>

### DIFF
--- a/contrib/kernel-install/91-sbctl.install
+++ b/contrib/kernel-install/91-sbctl.install
@@ -28,6 +28,14 @@ fi
 case "$COMMAND" in
 add)
 	printf 'sbctl: Signing kernel %s\n' "$IMAGE_FILE"
+
+	# exit without error if keys don't exist
+	# https://github.com/Foxboron/sbctl/issues/187
+	if ! test -d /usr/share/secureboot/keys; then
+		echo "Secureboot key directory doesn't exist, not signing!"
+		exit 0
+	fi
+
 	sbctl sign -s "$IMAGE_FILE" 1>/dev/null
 	;;
 remove)


contrib/kernel-install/91-sbctl.install
```diff
@@ -27,6 +27,14 @@ case "$COMMAND" in
 add)
 	[ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] &&
 		printf 'Signing kernel %s\n' "$IMAGE_FILE"
+
+	# exit without error if keys don't exist
+    # https://github.com/Foxboron/sbctl/issues/187
+	if ! test -d /usr/share/secureboot; then
```
Foxboron, line 33: `sbctl --json status | grep installed | awk '{print $2}' | tr -d ','`

or something like this would probably be a better idea?

ajakk: Seems like that doesn't always output something when the keys exist?

```
# sbctl create-keys
Created Owner UUID e8be777f-2cc7-4da5-bcfa-36fd80241bca
Creating secure boot keys...✓
Secure boot keys created!
c1ef6e49abfa / # sbctl --json status | grep installed | awk '{print $2}' | tr -d ','
c1ef6e49abfa / # sbctl --json status
```

Foxboron: Uhhhhhh, that is a bug :)

ajakk: I can reproduce like:

```
$ docker run --volume ~/git/sbctl:/sbctl -it archlinux bash
[root@f5c510c0e4a7 sbctl]# pacman --noconfirm --noprogressbar -Syu
[snip]
[root@f5c510c0e4a7 sbctl]# pacman --noconfirm --noprogressbar -S make go asciidoc gcc git
[snip]
[root@f5c510c0e4a7 sbctl]# go build -buildvcs=false -o sbctl ./cmd/sbctl
[snip]
[root@f5c510c0e4a7 sbctl]# ./sbctl create-keys
Created Owner UUID e0812b04-a6dd-4344-99a1-4691f4c20e07
Creating secure boot keys...✓
Secure boot keys created!
[root@f5c510c0e4a7 sbctl]# ./sbctl --json status
```

.. where ~/git/sbctl is this repository, updated to dcdc703, latest at the time of writing

Foxboron: Right, the docker container has access to efivars? Alternatively, do you get anything dropping `--json`?

ajakk: Ah, yes:

```
 ./sbctl status
system is not booted with UEFI
```

Which makes sense, because the container is indeed not booted with UEFI. So in this way, `sbctl status` seems unsuitable as a condition for signing, because signing *does* work after only doing `sbctl create-keys` in this environment.

ajakk: I agree, thanks!

ajakk:
Friendly ping! :)

Foxboron:
Sorry, been busy with the lower layers in `go-uefi` so I can test more easily. That involves solving this as I have virtual FS support to cover all the test cases :)

Foxboron (approved):

Foxboron:
Thanks!

Foxboron:
@ajakk now it's properly fixed... :)

https://github.com/Foxboron/sbctl/pull/323/commits/c4d803a9d3fd21efa9d1af0d96644fb4c05ffe42

behrmann:
If you intend to add more things after the `install` in the JSON this will break because of the trailing comma and it should probably be something like
```bash
sbctl setup --print-state --json | awk '/installed/ { gsub(/,$/,"",$2); print $2 }'
```

Should I open a PR or is the state done and will not be extended?

Foxboron:
> Should I open a PR or is the state done and will not be extended?

Isn't this trivially solved by just globbing it? 

```
[ "$(sbctl setup --print-state --json | grep installed | awk '{print $2}')" = *"true"* ];
```

or something

behrmann:
In dash I get
```
$ [ 'true,' = *"true"* ]; echo $?
1
```

Foxboron:
More awk it is then :) Please submit a PR when you have the time.

Issue #181, opened 2022-12-11
MSI PRO Z790-A has very broken Secure Boot


dawidpotocki:
Submitting as Foxboron asked me to.
> Feel free to open an issue on sbctl and we can keep track of it

Chat log: https://view.matrix.org/room/!GtIfdsfQtQIgbQSxwJ:archlinux.org/?anchor=$smbsrKoEbpAzdUlEfGDSA-StHWVsoVbotfH06e3lVXs&offset=120

The board seems to accept unsigned OSes when booting with Secure Boot
enabled. Firmware version: E7E07IMS.A22 (7E07vA2).

This is how it went:

I went to firmware to enable Secure Boot and also wiped factory keys to
switch into Setup Mode. Then I checked `sbctl status` and noticed that
while Secure Boot was enabled, Setup Mode was disabled and Microsoft was
still enrolled.

```
$ sbctl status
Installed:      ✓ sbctl is installed
Owner GUID:     6e2b1be3-6ecf-41e6-a59b-a30efdfa85ba
Setup Mode:     ✓ Disabled
Secure Boot:    ✓ Enabled
Vendor Keys:    microsoft
```

Well that made no sense. I went back to firmware and toggled "Custom"
mode to "Standard" or something like that. Well… still booting.

Here is the output of commands that I was asked to run:

```
$ efi-readvar -v db
Variable db, length 3961
db: List 0, type X509
    Signature 0, size 790, owner 841d6484-c82e-4135-ac79-0bf5d26284cd
        Subject:
            CN=MSI SHIP DB
        Issuer:
            CN=MSI SHIP KEK
db: List 1, type X509
    Signature 0, size 1572, owner 77fa9abd-0359-4d32-bd60-28f4e78f784b
        Subject:
            C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Corporation UEFI CA 2011
        Issuer:
            C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Corporation Third Party Marketplace Root
db: List 2, type X509
    Signature 0, size 1515, owner 77fa9abd-0359-4d32-bd60-28f4e78f784b
        Subject:
            C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Windows Production PCA 2011
        Issuer:
            C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Root Certificate Authority 2010
```

```
[...]

PR #189, opened 2023-01-11
status: Warn about firmware quirks

dawidpotocki:
Related: https://github.com/Foxboron/sbctl/issues/181

Just a rough idea, would like to hear your input.

sbctl status output

```
{
  "installed": true,
  "guid": "6e2b1be3-6ecf-41e6-a59b-a30efdfa85ba",
  "setup_mode": false,
  "secure_boot": true,
  "vendors": [],
  "firmware_quirks": [
    {
      "id": "FQ0001",
      "name": "MSI defaults to executing on policy violation",
      "link": "https://github.com/Foxboron/sbctl/wiki/FQ0001",
      "severity": "CRITICAL"
    }
  ]
}
```

### DIFF
--- a/cmd/sbctl/status.go
+++ b/cmd/sbctl/status.go
@@ -9,6 +9,7 @@ import (
 	"github.com/foxboron/sbctl"
 	"github.com/foxboron/sbctl/fs"
 	"github.com/foxboron/sbctl/logging"
+	"github.com/foxboron/sbctl/quirks"
 	"github.com/spf13/cobra"
 )
 
@@ -19,20 +20,22 @@ var statusCmd = &cobra.Command{
 }
 
 type Status struct {
-	Installed  bool     `json:"installed"`
-	GUID       string   `json:"guid"`
-	SetupMode  bool     `json:"setup_mode"`
-	SecureBoot bool     `json:"secure_boot"`
-	Vendors    []string `json:"vendors"`
+	Installed      bool           `json:"installed"`
+	GUID           string         `json:"guid"`
+	SetupMode      bool           `json:"setup_mode"`
+	SecureBoot     bool           `json:"secure_boot"`
+	Vendors        []string       `json:"vendors"`
+	FirmwareQuirks []quirks.Quirk `json:"firmware_quirks"`
 }
 
 func NewStatus() *Status {
 	return &Status{
-		Installed:  false,
-		GUID:       "",
-		SetupMode:  false,
-		SecureBoot: false,
-		Vendors:    []string{},
+		Installed:      false,
+		GUID:           "",
+		SetupMode:      false,
+		SecureBoot:     false,
+		Vendors:        []string{},
+		FirmwareQuirks: []quirks.Quirk{},
 	}
 }
 
@@ -67,6 +70,13 @@ func PrintStatus(s *Status) {
 	} else {
 		logging.Println("none")
 	}
+	if len(s.FirmwareQuirks) > 0 {
+		logging.Print("Firmware:\t")
+		logging.Print(logging.Warnf("Your firmware has known quirks"))
+		for _, quirk := range s.FirmwareQuirks {
+			logging.Println("\t\t- " + quirk.ID + ": " + quirk.Name + " (" + quirk.Severity + ")\n\t\t  " + quirk.Link)
+		}
+	}
 }
 
 func RunStatus(cmd *cobra.Command, args []string) error {
@@ -90,6 +100,7 @@ func RunStatus(cmd *cobra.Command, args []string) error {
 	if keys := sbctl.GetEnrolledVendorCerts(); len(keys) > 0 {
 		stat.Vendors = keys
 	}
+	stat.FirmwareQuirks = quirks.CheckFirmwareQuirks()
 	if cmdOptions.JsonOutput {
 		if err := JsonOut(stat); err != nil {
 			return err
--- a/cmd/sbctl/status_test.go
+++ b/cmd/sbctl/status_test.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"reflect"
 	"testing"
+	"testing/fstest"
 
 	"github.com/foxboron/go-uefi/efi/efitest"
+	"github.com/foxboron/sbctl/quirks"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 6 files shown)


Foxboron:
I like this!

Generally I would need help to maintain a list of quirks, we probably need a new section as well.

- `/sys/devices/virtual/dmi/id/` reading should be serialized into a struct. Probably it's own module as well since we might want to use this for debug output or an `status --expert` thing.
- I think we can express [fq0001.go](https://github.com/Foxboron/sbctl/pull/189/files#diff-ad211c5071c23f8686862325f45c4f99f163f1b8a7d8508a9bac3d1bfdf55e1a) entierly as a structs if the above is done.


Generally it's a very good first draft :) Just need to figure out how to best document all of this!

dawidpotocki:
I found that it is possible to get the IFR data from the OS.

This method relies on reading HiiDB efivar which has the offset and size
of the UEFI Human Interface Infrastructure that we can use to read data
we need from memory.

The only issue is that it requires reading from `/dev/mem` and thus root
access. Because of it, it won't work if Kernel Lockdown is enabled in
confidentiality mode, but it should work in integrity mode. Another
thing is that it doesn't tell us if the user has it enabled, it only
tells us the default values.

I wrote a very messy proof of concept in half Python, half shell:

```py
#!/usr/bin/env python3

import os, shutil, struct

with open("/sys/firmware/efi/efivars/HiiDB-1b838190-4625-4ead-abc9-cd5e6af18fe0", "rb") as f:
    size, addr = struct.unpack('@xxxxII', f.read())
with open("/dev/mem", "rb") as f:
    f.seek(addr)
    hiidb = f.read(size)

if hiidb.replace(b"\x00",b"").find(b"Image Execution Policy") == -1:
    print("Image Execution Policy settings not found in HiiDB")
else:
    print("Image Execution Policy settings found in HiiDB")
    try:
        shutil.rmtree("/tmp/hiidb")
    except FileNotFoundError:
        pass
    os.mkdir("/tmp/hiidb")
    with open("/tmp/hiidb/hiidb", "wb") as f:
        f.write(hiidb)
	os.system("""
		cd /tmp/hiidb
		ifrextractor /tmp/hiidb/hiidb 1>/dev/null
		output="$(grep -A1 -E 'OneOf Prompt: "(Option ROM|Removable Media|Fixed Media)", Help: "Image Execution Policy' /tmp/hiidb/*ifr.txt)"

		if echo "$output" | grep -q "DefaultId: 0x0"; then
			printf "\033[1;31mInsecure defaults\033[0m\n" "$1"
		else
			printf "\033[1;32mSecure defaults\033[0m\n" "$1"
		fi
	""")
    shutil.rmtree("/tmp/hiidb")
```

Because of the issues with this approach, I don't think it would make
sense to add anything like this to sbctl, even though it would be more
accurate than checking dates of firmware.

I don't know how else I would be able to read this data if not from
`/dev/mem`. Suggestions are welcome.

Foxboron:
I've started improving the end-to-end unit testing support in `sbctl`.

So if you want to try write unit tests for the CLI interface to test this feature it should be possible :)


https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/status_test.go

https://github.com/Foxboron/go-uefi/blob/master/efi/efitest/files.go

dawidpotocki:
Sure thing, will do.

I just had some free time and tried, but I'm not really sure how I'm
supposed to set files.

```go
SetFS(
	fstest.MapFS{
		"/sys/devices/virtual/dmi/id/board_vendor": {
			Data: []byte{0x0, 0x6, 0x0, 0x0, 0x1}},
	},
	fstest.MapFS{
		"/sys/firmware/efi/efivars/SetupMode-8be4df61-93ca-11d2-aa0d-00e098032b8c": {
			Data: []byte{0x0, 0x6, 0x0, 0x0, 0x1}},
	},
	efitest.SecureBootOn(),
)
```

I tried something like this and I get such output:

```
{false  true true [] [{FQ0001 Defaults to executing on Secure Boot policy violation https://github.com/Foxboron/sbctl/wiki/FQ0001 CRITICAL date}]}
```

When I change Setup Mode part to

```
Data: []byte{0x0, 0x6, 0x0, 0x0, 0x0}},
```
```
{false  false true [] [{FQ0001 Defaults to executing on Secure Boot policy violation https://github.com/Foxboron/sbctl/wiki/FQ0001 CRITICAL date}]}
```

When I change the check in fq0001.go

```go
if dmi.Table.BoardVendor == "not Micro-Star International Co., Ltd." && dmi.Table.ChassisType == "3" {
```
```
{false  true true [] []}
```

So clearly it does set Setup Mode and Secure Boot efivar files, but it
doesn't change the `board_vendor` file at all. I assume I'm doing
something very wrong, but not sure what.

Will look more into it in the morning in my timezone.

Foxboron:
Keep in mind you need to shim all filesystem calls with the `fs` module :)

https://github.com/Foxboron/sbctl/commit/e1ce225d880c1db11bbd4b73f8d969d63c85ea3f
https://github.com/Foxboron/sbctl/commit/57f51d9c7ef962aab29166c7adf17c5af16e2418

dawidpotocki:
I had actually tried changing `os` to `sbctl/fs`… and I just now
realised what was the problem.

`dmi.go` had `var Table = parseDMI()`, which I guess was ran before the
`SetFS()` stuff.

If we add `dmi.ParseDMI()` inside `fq0001.go`'s `HasFQ0001()`, it works.

Idk what to do with this now.

Foxboron:
>Idk what to do with this now, I guess I can just execute it in quirks.go.

Please only parse it when we need it :) We probably don't need this as global state

dawidpotocki:
Since there is no way to PR to wiki, here are my changes:
https://github.com/dawidpotocki/sbctl/wiki

Foxboron:
I did buy the domain `secureboot.dev` a while ago :)

https://github.com/Foxboron/secureboot.dev

Might be an idea to figure out a usecase for that thing.

Foxboron (approved):

Foxboron:
Thanks for working on this :)

jdholtz:
Thanks for adding this feature! I just applied the changes from the [wiki](https://github.com/Foxboron/sbctl/wiki/FQ0001) and it worked from my testing. Will these quirks still be displayed even after the changes in the BIOS are made? I want to make sure I have it set up properly and didn't know if this warning was still telling me otherwise.

dawidpotocki:
Yes, they will still be displayed, at least this one.
This is because it's hard to reliably detect this setting at runtime.

PR #192, opened 2023-01-14
enroll-keys: Enroll Microsoft KEK along with their other keys

alois31:
If enrolling the Microsoft db keys, also enroll their KEK. This allows applying the official dbx updates, which are important for security, as they allow blocking binaries with known vulnerabilities which Microsoft has signed in the past.


Foxboron:
Nice, this implements what I had in mind to include the `KEK` key. Thanks.

I'll try and get to testing this and give some feedback soon'ish :) Just have a long todo atm.

Foxboron (approved):

PR #194, opened 2023-01-20
Fix arbitrary sizes in UKI generation

eNV25:
I noticed UKI generation in sbctl uses hardcoded values even though the ArchWiki advises against it, and mkinitcpio was updated to fix this.

The logic is based on the ArchWiki and mkinitcpio implementations, but uses the `debug/pe` package to calculate the VMA/offset, instead of objdump+awk.

https://wiki.archlinux.org/title/Unified_kernel_image#Manually
https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/182
https://pkg.go.dev/debug/pe

### DIFF
--- a/bundles.go
+++ b/bundles.go
@@ -1,6 +1,7 @@
 package sbctl
 
 import (
+	"debug/pe"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Foxboron:
Awesome, thanks :)

All this code was written before people starting discussing the size contraints of the original objcopy line.

Foxboron (approved):

bundles.go
```diff
@@ -127,15 +128,49 @@ func NewBundle() (bundle *Bundle, err error) {
...
+	sections := []section{
+		{".osrel", bundle.OSRelease},
+		{".cmdline", bundle.Cmdline},
+		{".splash", bundle.Splash},
+		{".linux", bundle.KernelImage},
+		{".initrd", bundle.Initramfs},
+	}
+
+	var args []string
+	vma, err := func() (int64, error) {
```
eNV25, line 144: @Foxboron 

systemd's `ukify` rounds up to block size (4096). Should we do the same?

https://github.com/systemd/systemd/blob/d09df6b94e0c4924ea7064c79ab0441f5aff469b/src/ukify/ukify.py#L80-L90

Foxboron: Unclear to me why. But yes please.

I also think that needs to be done in `mkinitcpio` when I think about it.

eNV25: Also, `ukify` uses `pe.OPTIONAL_HEADER.ImageBase + section.VirtualAddress + section.Misc_VirtualSize` as the base offset.

I'll see if there is something like ImageBase in [debug/pe]

medhefgo:
Next systemd release has a `ukify` tools that's (hopefully) going to be the only supported (by upstream) means of assembling UKIs. You should consider switching to that instead.

eNV25:
That's a good idea. If `ukify` is installed, we use that, or else fall back to `objcopy`.

medhefgo:
> That's a good idea. If `ukify` is installed, we use that, or else fall back to `objcopy`.

Use llvm-objcopy. Supports all arches and doesn't need manual VMA calculation (nor does it support it afaik).

eNV25:
I don't think `llvm-objcopy` works.

I tried it on my machine, locally editing `sbctl` to use `llvm-objcopy` and skipping `--change-section-vma`.

Running `bootctl status` gets the following message:

```
Missing fields in os-release data from unified kernel image /efi/EFI/Linux/archboot.efi, refusing.
```

Running `objdump -h` reports that the new sections have 0 as VMA and LMA.

medhefgo:
We had the same issue; you need to pass `--set-section-flags` to mark each section you add as `data,readonly` or `code,readonly` (for `.linux`) to make it work.

eNV25:
I got that working.

I guess that's an option as well if we can depend on llvm.

Foxboron:
@eNV25 are you happy with the state of this? I see there are a few functions we should probably add to `go-uefi` in the future but generally this seems fine from my end :)

eNV25:
Yes, this looks good to me.

medhefgo:
I take it you don't want to at least move to llvm-objcopy?

eNV25:
That's fine as well, but that would mean a new dependency. @Foxboron should make that decision.

Foxboron:
@medhefgo I don't feel like having llvm as a dependency :) There doesnt seem like there is a good reason to switch?

bundles.go
```diff
@@ -126,16 +127,64 @@ func NewBundle() (bundle *Bundle, err error) {
...
-		"--add-section", fmt.Sprintf(".initrd=%s", bundle.Initramfs), "--change-section-vma", ".initrd=0x3000000",
+	type section struct {
+		section string
+		file    string
+	}
+	sections := []section{
+		{".osrel", bundle.OSRelease},
+		{".cmdline", bundle.Cmdline},
+		{".splash", bundle.Splash},
+		{".linux", bundle.KernelImage},
```
medhefgo, line 142: The kernel needs a buffer area after it, or there is a chance it clobbers the initrd when run through efi handover protocol. In systemd we add it last to hopefully avoid this case. (This is only for old kernels.)

medhefgo:
The best reason would be objcopy only ever supports a limited set of arches (you cannot prepare an arm image on x86 with objcopy, for example) , while llvm handles them all.

Long term I still recommend using ukify from v253 and sidestep the matter…

Foxboron:
Right, I feel like dropping UKI support from `sbctl`. I considered it mostly a tacked on feature because most distros where lacking in proper tooling for it.

It can stay until it lags behind the general stuff like `mkinitcpio`, `dracut` and `ukify`.

bundles.go
eNV25: How do we fix that? How much buffer space does it need?

Also, wouldn't it make more sense to keep `.linux` after `.initrd`?

Foxboron:
@eNV25 sorry, git migration and FOSDEM stuff has taken up time. I'll do review stuff and a release this week :)

PR #217, opened 2023-05-11
Allow enrolling custom db and KEK certs

### LINKED ISSUE #190: Add support for custom OEM keys
This is a feature request to support enrolling custom SecureBoot DB certificates in addition to the current oem certificates given by the enroll options `--microsoft` and `--tpm-eventlog`.
I think this could be implemented by adding a new OEM, which stores it's certificates in a subdirectory of `/usr/share/secureboot/`

My direct usecase would be for building custom kernel modules and signing them, e.g. for DKMS enrolling `/var/lib/dkms/mok.pub` in Debian.

Cornelicorn:
Closes #190.

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -3,6 +3,7 @@ package certs
 import (
 	"embed"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/foxboron/go-uefi/efi/signature"
@@ -17,6 +18,7 @@ var (
 	oemGUID      = map[string]util.EFIGUID{
 		"microsoft":    *util.StringToGUID("77fa9abd-0359-4d32-bd60-28f4e78f784b"),
 		"tpm-eventlog": *util.StringToGUID("4f52704f-494d-41736e-6e6f79696e6721"),
+		"custom":       *util.StringToGUID("88a69775-5ad7-45d9-9f34-cec43e1f1989"),
 	}
 )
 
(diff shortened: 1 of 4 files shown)


Cornelicorn:
The hardcoded Owner UUID is just a placeholder.
I'm not sure if the Owner UUID should be dynamically associated with every certificate, set once for all custom certificates (statically in sbctl or read from the filesystem as the certificates) or the existing UUID (`/usr/share/secureboot/GUID`) should be reused

Foxboron:
I think utilizing a custom GUID for this is fine. It makes it easier to make assumptions on when looking at the signature database.

Cornelicorn:
> I think utilizing a custom GUID for this is fine. It makes it easier to make assumptions on when looking at the signature database.

Ok, I put in a newly generated UUID (instead of the tpm-eventlog one incremented by 1).

Foxboron:
If you want an additional challenge, you could try to write a test for this using the virtual filesystem layer `sbctl` has :)

Example: https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/status_test.go

If you don't have time that is fine as well, but getting more test coverage would be cool.

Cornelicorn:
You mean a test to check that enrolling custom certificates works?

Foxboron:
Yep :)

Cornelicorn:
After looking through the testing suite, I don't think that's quite doable for me right now (It's the first time for me to work with go), sorry.

Foxboron:
That is fine :) No worries!

Foxboron (approved):

Cornelicorn:
I think GitHub still wants a separate approval for the workflow runs, besides the PR review.

Foxboron:
Thanks for implementing this :)

PR #222, opened 2023-06-16
Add support for loading certificates from dbDefault

Foxboron:
Supersedes https://github.com/Foxboron/sbctl/pull/147

### DIFF
--- a/certs/builtin.go
+++ b/certs/builtin.go
@@ -1,0 +1,71 @@
+package certs
+
+import (
+	"fmt"
+	"os"
+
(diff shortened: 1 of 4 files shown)


DHowett:
This is great, and I love it. Thanks for picking up the torch I dropped!

DHowett:
I may just be saying this because I officially and technically work at a big company, but: I know that open-source licensing can be a pain, and I would be happy to sign a CLA or a copyright transfer agreement. That will make me one less person you will need to track down if you ever choose to relicense sbctl :smile:

PR #231, opened 2023-07-31
feat(enroll-keys): add partial enrollment of keys

RiSKeD:
Allow keys to be partially enrolled via --partial flag.

### DIFF
--- a/cmd/sbctl/enroll-keys.go
+++ b/cmd/sbctl/enroll-keys.go
@@ -42,7 +42,15 @@ func (s *StringSet) Set(p string) error {
 }
 
 func (s *StringSet) Type() string {
-	return "[auth, esl]"
+	var allowedValues string
+
+	for _, allowedValue := range s.Allowed {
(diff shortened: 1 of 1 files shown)


Foxboron:
Is it possible to get the use-cased described?

RiSKeD:
We intend to use sbctl in some automated test cases for SecureBoot on a few platforms (key enrollment, resets, and rotation): 
To facilitate this, it would be nice to rollout/reset/rotate each key individually, so we can separate the tests for each hierarchy.
This will also include importing and using invalid keys besides valid ones in the future.

Foxboron (approved):

Foxboron:
Sure, that makes sense. LGTM.

Thanks :)

RiSKeD:
Could you merge the changes? I got another PR in the pipeline based on these changes :smile:

Foxboron:
> Could you merge the changes? I got another PR in the pipeline based on these changes smile

Sorry, not all caught up after vacation so some ramp-up time expected :)

PR #234, opened 2023-08-07
Partial rotate

RiSKeD:
1. Instead of creating keys directly, the user can now supply their own keys via --new-keys-dir.
2. The user can do a partial rotation for each hierarchy by providing a key + cert file.


RiSKeD:
Part of #236

PR #236, opened 2023-08-09
Dbx Key Management

RiSKeD:
Enrollment/Rotation/Reset of dbx keys.


Foxboron (approved):

Foxboron:
Thanks!

Foxboron:
@RiSKeD So, I should have caught this during review but the `dbx` implementation here doesn't *really* make sense.

Creating a key and expecting it to be enrolled is not really how we use `dbx` in UEFI, it should be a revocation list and be specially handled. Currently it's breaking key enrollment for `enroll-keys`.

I'll need to partially revert this patch, but if you can explain how you are currently using this feature I can try and not break it for your usecase.

PR #239, opened 2023-08-15
feat(custom-keys): roll out any bytes to the specificed hierarchy

RiSKeD:
Use-Case: Testing that SecureBoot keys with a invalid signature or signing date must not be added to the efi.


Foxboron:
We already have `enroll-keys --custom`, can we extend this instead of introducing a new command?

RiSKeD:
> We already have `enroll-keys --custom`, can we extend this instead of introducing a new command?

Sure, I was not sure if that would be best as the command already has a bunch of different options, but I can incorporate the functionality into it :+1:

RiSKeD:
> I might be missing something, but this is not really useful for general use of wanting to manage valid secureboot keys and should be possible using `efivar(8)`, no?

In general, I agree with you, but i would like to put one additional flag inside the enroll-keys command to roll out invalid/empty/... keys for testing purposes. The convenience of sbctl as a pure-go tool is beneficial for some environment.

Foxboron:
semi-forgot this in the middle of work and camp :upside_down_face: 

Generally I don't mind adding specialized flags to `sbctl`. We can hide them by default or introduce an `--expert` command if it would make things easier for us.

Foxboron:
Generally, I think I should spend more time fixing the plumbing inside `go-uefi` so we can utilize the integration/end-to-end tests that I started working on. It would make it easier to introduce features like this and also have them testable.

https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/status_test.go

Otherwise, LGTM.

Foxboron (approved):

Foxboron:
Thanks!

Issue #167, opened 2022-10-05
Please provide a package for Ubuntu 22.04 LTS


IPlayZed:
Hi!

It would be nice if you could package (or provide build instructions) for Ubuntu 22.04 LTS.

Foxboron:
I don't really want to learn deb packaging so someone would need to provide this.

IPlayZed:
How do I build on Arch?

I would be willing to look into packaging it for deb if I know a bit more.

klausenbusk:
Please check out the PKGBUILD: https://github.com/archlinux/svntogit-community/blob/packages/sbctl/trunk/PKGBUILD.

dngray:
Maybe someone might like to take a look at https://wiki.debian.org/RFP. I'm currently working on getting this into Fedora.

IPlayZed:
It would be great if this was streamlined across distros, it is only tool providing a good implementation for secure boot management imho. The problem with shims is that they do not solve the problem of unsigned initramfs images, which result in a lax interpretation of what is trusted, so if you want to be 100% compliant in corporate, you have to make sure to generate your own EFI bundles and sign them manually. This tool provides that.

Foxboron:
People are free to contribute packages to their respective distributions, but this is not something I'll be pushing for personally as I lack the time for it.

For a list of existing packages please see repology. https://repology.org/project/sbctl/versions

Could probably mention this on the README as well.

IPlayZed:
Please don't misunderstand my comment @Foxboron  I am not pushing you personally, this is just a discussion.

Foxboron:
No worries :) It's just a general statement from my side.

sbskas:
As a matter of fact, when I need to be cross-compatible with different distro, I use fpm to convert packages from one package format to another one.
It may be of some help to automically convert packages or to generate basic packages skeletons.

jloeser:
I needed it too and built it on openSUSE Build Service.

You can install it via:
```
root@vm:~# . /etc/os-release
root@vm:~# wget https://download.opensuse.org/repositories/home:/jloeser:/secureboot/xUbuntu_$VERSION_ID/Release.key -O- | apt-key add -
root@vm:~# add-apt-repository -y "deb https://download.opensuse.org/repositories/home:/jloeser:/secureboot/xUbuntu_$VERSION_ID/ ./"
root@vm:~# apt update
root@vm:~# apt install sbctl
root@vm:~# sbctl status
[...]

PR #242, opened 2023-09-06
Add Debian repository information to README

julianfairfax:
I have created a Debian package for sbctl which is available in my Debian repository.

In the future, I may look into getting this tool into the official Debian repos, but that is not as easy as creating this package was.

I also spent a lot of time trying to create something similar to the pacman hook, for automatic resigning after a modification of Debian's EFI files, but I was unsuccessful.

I don't have time to further investigate that, but, if someone knows how to do this, within the framework of the [package setup](https://gitlab.com/julianfairfax/package-repo/-/blob/main/sbctl/build_deb.sh?ref_type=heads), then I will implement it.

In any case, hopefully this package will help users who want to use sbctl on Debian, but don't feel like compiling it. I can tell you sbctl works the same way on Debian as it does on Arch, despite it not being documented on the Debian wiki. Thanks to sbctl, I have working secure boot on Debian on my laptop (which didn't come with Microsoft's keys)!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,14 @@ For Alpine Linux:
 $ apk add sbctl
 ```
 
+For Debian Linux:
+
+Add [Julian's repository](https://gitlab.com/julianfairfax/package-repo#how-to-add-repository-for-debian-based-linux-distributions), then run:
+```
+$ sudo apt update # after adding the repository
+$ sudo apt install sbctl
+```
+
 You can find a updated list of [sbctl packages on
 Repology](https://repology.org/project/sbctl/versions).
 


julianfairfax:
This should close #167

julianfairfax:
Could someone take a look at this please?

Foxboron:
I have looked over it and I'm not quite confident in the packaging work to include it in the `README.md`.

Just throwing the binary into `/usr/bin` and including a minimal metadata file, that doesn't even include a license, doesn't seem right.

There is an entire `make install` command that is just not being used and I'm not sure if scripts like these are the proper way to package Debian packages.

julianfairfax:
> Just throwing the binary into /usr/bin and including a minimal metadata file, that doesn't even include a license, doesn't seem right.

> There is an entire make install command that is just not being used

I am now using that, and all expected files are now included.

> and I'm not sure if scripts like these are the proper way to package Debian packages.

They are a way to make Debian packages. If by "proper way", you mean the way you would have to do it for the package to by accepted into the Debian repos, then this is not that way.

However, creating such a Debian package is also requires packaging all dependencies. As this tool is written in go, which is a newer programming language and has less packages in Debian, that could be a long and complicated process, especially since each dependency of sbctl that is not packaged in Debian, could also have its own dependencies that also require packaging.

This is something I would like to do at some point. But at the moment I don't have the time or the willingness to tackle that task. This package, that I have created, is exactly the same as how a "proper" Debian package would be, except for its creation method.

Additionally, this is a more interesting improvement to me than the creation process of the package:

> I also spent a lot of time trying to create something similar to the pacman hook, for automatic resigning after a modification of Debian's EFI files, but I was unsuccessful.

> I don't have time to further investigate that, but, if someone knows how to do this, within the framework of the [package setup](https://gitlab.com/julianfairfax/package-repo/-/blob/main/sbctl/build_deb.sh?ref_type=heads), then I will implement it.

Foxboron:
I don't intend for you to package all the dependencies. But having a `build_deb.sh` just seems like a poor way to package this.

julianfairfax:
> I don't intend for you to package all the dependencies. But having a `build_deb.sh` just seems like a poor way to package this.

There is no alternative. The only other way of building the package is the "proper" Debian way, which requires building all the dependencies.

I'm not sure what it is that you do want, but if this package isn't to your liking, then you should close the PR. The fact is that the package does work exactly as a Debian package should.

```
user@verbose:~$ apt info sbctl
Package: sbctl
Version: 0.11
Priority: extra
Section: misc
Maintainer: Julian <juliannfairfax@protonmail.com>
Installed-Size: unknown
Download-Size: 4,145 kB
APT-Manual-Installed: yes
APT-Sources: https://julianfairfax.gitlab.io/package-repo/debs packages/main amd64 Packages
Description: Secure Boot key manager

user@verbose:~$ dpkg-query -L sbctl
/.
/usr
/usr/bin
/usr/bin/sbctl
/usr/lib
/usr/lib/kernel
/usr/lib/kernel/install.d
/usr/lib/kernel/install.d/91-sbctl.install
/usr/share
/usr/share/bash-completion
/usr/share/bash-completion/completions
/usr/share/bash-completion/completions/sbctl
/usr/share/fish
/usr/share/fish/vendor_completions.d
/usr/share/fish/vendor_completions.d/sbctl.fish
/usr/share/licenses
/usr/share/licenses/sbctl
/usr/share/licenses/sbctl/LICENSE
/usr/share/man
/usr/share/man/man8
/usr/share/man/man8/sbctl.8
/usr/share/zsh
/usr/share/zsh/site-functions
/usr/share/zsh/site-functions/_sbctl
```

PR #244, opened 2023-09-14
Add append option

RiSKeD:
Added one more flags to enroll-keys and reset respectively: 
Enroll-keys: Append the key to be enrolled to the existing db
Reset: Only remove a specific certs by providing a flag with their paths

I am hesitant to add more flags to the enroll-keys cmd but they do seem useful to me

### DIFF
--- a/cmd/sbctl/reset.go
+++ b/cmd/sbctl/reset.go
@@ -4,18 +4,22 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/foxboron/go-uefi/efi"
+	"github.com/foxboron/go-uefi/efi/signature"
 	"github.com/foxboron/go-uefi/efi/util"
 	"github.com/foxboron/sbctl"
+	"github.com/foxboron/sbctl/fs"
 	"github.com/foxboron/sbctl/logging"
 	"github.com/foxboron/sbctl/stringset"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


cmd/sbctl/reset.go
```diff
@@ -150,6 +207,7 @@ func RunReset(cmd *cobra.Command, args []string) error {
 func resetKeysCmdFlags(cmd *cobra.Command) {
 	f := cmd.Flags()
 	f.VarPF(&resetCmdOpts.Partial, "partial", "p", "reset a partial set of keys")
+	f.StringVarP(&resetCmdOpts.CertFiles, "cert-files", "c", "", "optional paths to certificate file to remove from db (seperate individual paths by ';')")
```
Cornelicorn, line 210: This may not only be db, but also other variables, right? Even though the function is called the same, this seems a bit confusing to me from the wording

RiSKeD: right, db is referring to the cert database of any hierarchy,  the wording is a bit off there.

Foxboron:
Missing a section in the manpage :)

https://github.com/Foxboron/sbctl/blob/master/docs/sbctl.8.txt#L35

RiSKeD:
> Missing a section in the manpage :)
> 
> https://github.com/Foxboron/sbctl/blob/master/docs/sbctl.8.txt#L35

I added comments for each of the flags I added.

Foxboron (approved):

PR #254, opened 2023-10-26
Check and return Open errs

quite:


Foxboron (approved):

quite:
Looks like build on Void is broken

Foxboron:
Seems like it's a TLS certificate error, so I'll just ignore it :) Thanks

Issue #102, opened 2021-10-18
linux refusing the creation of PK on thinkpad T460s


dkwo:
I'm trying to use `sbctl` on Void linux, but it appears I'm not mounting the efivars correctly.
Could you please point out the correct way to do it, without relying on dbus, udev etc?
```
$ doas sbctl status
Installed:      ✔ Sbctl is installed
Owner GUID:     <some stuff>
Setup Mode:     ✔ Disabled
Secure Boot:    ✘ Disabled
$ doas sbctl enroll-keys
Enrolling keys to EFI variables...✘
couldn't sync keys: couldn't open file: open /sys/firmware/efi/efivars/<some string>: no such file or directory
$ doas mount -t efivarfs efivarfs /sys/firmware/efi/efivars
$ doas sbctl status
Installed:      ✔ Sbctl is installed
Owner GUID:     <same stuff>
Setup Mode:     ✘ Enabled
Secure Boot:    ✘ Disabled
$ doas sbctl enroll-keys
Enrolling keys to EFI variables...✘
sbtl requires root to run: couldn't sync keys: couldn't write efi variable: write /sys/firmware/efi/efivars/<some different string>: permission denied
```
Thanks.

Foxboron:
Output of `ls -lah /sys/firmware/efi/efivars/PK-8be4df61-93ca-11d2-aa0d-00e098032b8c /sys/firmware/efi/efivars/KEK-8be4df61-93ca-11d2-aa0d-00e098032b8c /sys/firmware/efi/efivars/db-d719b2cb-3d3a-4596-a3bc-dad00e67656f` please

dkwo:
```
 ls -lah /sys/firmware/efi/efivars/PK-8be4df61-93ca-11d2-aa0d-00e098032b8c /sys/firmware/efi/efivars/KEK-8be4df61-93ca-11d2-aa0d-00e098
032b8c /sys/firmware/efi/efivars/db-d719b2cb-3d3a-4596-a3bc-dad00e67656f
-rw-r--r-- 1 root root 1.4K Oct 18 19:12 /sys/firmware/efi/efivars/KEK-8be4df61-93ca-11d2-aa0d-00e098032b8c
-rw-r--r-- 1 root root    0 Oct 18 19:12 /sys/firmware/efi/efivars/PK-8be4df61-93ca-11d2-aa0d-00e098032b8c
[...]

PR #255, opened 2023-11-01
keys.go: drop the keyUsage bitfield

dkwo:
fixes https://github.com/Foxboron/sbctl/issues/102

been testing this for a while on my thinkpad t460s

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -59,7 +59,6 @@ func CreateKey(name string) ([]byte, []byte, error) {
 		SignatureAlgorithm: x509.SHA256WithRSA,
 		NotBefore:          time.Now(),
 		NotAfter:           time.Now().AddDate(5, 0, 0),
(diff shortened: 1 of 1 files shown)


Foxboron (approved):

dkwo:
btw, i also opened PR for void, to use correct go build style (pie, etc), update and include this patch, in case you want to take a look https://github.com/void-linux/void-packages/pull/46165

Cornelicorn:
There is another location with `KeyUsage`: https://github.com/Foxboron/sbctl/blob/master/tests/utils/certs.go#L27

dkwo:
@Cornelicorn Feel free to open a PR, if you believe that needs to be fixed as well.

Foxboron:
> There is another location with KeyUsage: https://github.com/Foxboron/sbctl/blob/master/tests/utils/certs.go#L27

Ah right, I was moving around the API at some point.

PR #261, opened 2023-11-13
tests/utils/certs.go: drop keyUsage bitfield

dkwo:
drop keyusage bitfield from another location, as pointed out in https://github.com/Foxboron/sbctl/pull/255


Foxboron:
Thanks!

PR #266, opened 2023-12-08
Update 91-sbctl.install

cvlc12:
Make it clear that messages are from sbctl, instead of them being lost in the kernel-install output.
Always say we're signing the kernel


Foxboron:
Generally looks good, but please squash the changes.

Foxboron:
Thanks!

PR #267, opened 2023-12-09
BUGFIX: f.StringVarP() was clearing the exportPath/databasePath strings

spillner:
I'm not sure how ```sbctl create``` ever worked with the default paths--- the var {} block initially sets exportPath/databasePath to the default values from the sbctl module, but then pflag.StringVarP() overwrote them with empty string passed in as the ```value``` parameter. From https://github.com/spf13/pflag/blob/master/string.go it looks like that has been the behavior for a long time.

Fixed by passing the values from sbctl as the initialization value in StringVarP().


Foxboron (approved):

Foxboron:
Thanks!

Issue #263, opened 2023-11-24
Include Target= for linux-firmware


duffydack:
Just had my arch update linux-firmware pkg, which triggered an initramfs rebuild, but left me without a bootable system as sbctl never got triggered. 

any chance of including 
`Target = usr/lib/firmware/*`
in your zz-sbctl.hook` ?

Foxboron:
That seems reasonable to me. You can also write a `mkinitcpio` post-hook that runs `sbctl`. Feel free to send a patch!

Issue #218, opened 2023-05-16
[RFE] Adding a mkinitcpio post generation hook


neitsab:
Hi, I just setup sbctl for the first time and noticed that manual invocation of `mkinitcpio` would not trigger a re-signing by sbctl. My use case is regenerating a UKI after updating the kernel cmdline.

The wiki for mkinitcpio describes a [post generation hook](https://man.archlinux.org/man/mkinitcpio.8#ABOUT_POST_HOOKS) for this use case[[1]](https://wiki.archlinux.org/title/Unified_kernel_image#Signing_the_UKIs_for_Secure_Boot), albeit one using sbsigntools.

Could that be something added to sbctl, like the pacman hook in the contrib/ folder? It's a shame I'm really not proficient at all in shell scripting, otherwise I would have provided one myself, but I'm really not sure what should be tested for etc.

Would you consider adding one to this great utility?
Thanks in any case!

**edit**: I'm on Arch, sbctl version 0.11-1, and other than that everything works fine, UKI signing on kernel install etc. UKI is generated by mkinitcpio and placed into `/boot/efi/EFI/Linux/arch-linux.efi`.

neitsab:
From my test, it seems that the following works:

```sh
cat << EOF > /etc/initcpio/post/kernel-sbctl
#!/usr/bin/env bash

kernel="$1"
[[ -n "$kernel" ]] || exit 0

# use already installed kernel if it exists
[[ ! -f "$KERNELDESTINATION" ]] || kernel="$KERNELDESTINATION"

sbctl sign "$kernel"
EOF
chmod +x /etc/initcpio/post/kernel-sbctl
```

(adapted from [the wiki](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot#Signing_the_kernel_with_a_mkinitcpio_post_hook))

Or it can be even as simple as:

```sh
cat << EOF > /etc/initcpio/post/uki-sbctl
#!/usr/bin/env bash
sbctl sign-all
EOF
chmod +x /etc/initcpio/post/uki-sbctl
```

if using a UKI with a stable name that has already been added to to sbctl signature database!

NekkoDroid:
you can also do this if you want to explicitly sign the UKIs if they were specified
```sh
#!/usr/bin/env bash
[[ -n "$3" ]] && sbctl sign "$3"
```

m00nwtchr:
[...]

PR #271, opened 2023-12-19
Ensure file signing hook is run when initrd is rebuilt

Joseph-DiGiovanni:
Initrd is rebuilt when a DKMS module or firmware package is installed or updated as of these commits to mkinitcpio: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/aff81712789b9f2c1664fe1cfb5c1ecdbc5c993b https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/3576b03d29420ccd1913eaa18c7f8950e7de3103

Without this change, images created by mkinitcpio will not be automatically signed for secure boot in the events mentioned above.

### DIFF
--- a/contrib/mkinitcpio/sbctl
+++ b/contrib/mkinitcpio/sbctl
@@ -1,3 +1,3 @@
-#!/usr/bin/env bash
+#!/usr/bin/bash
 echo "Signing EFI binaries..."
 /usr/bin/sbctl sign-all -g
--- a/contrib/mkinitcpio/sbctl
+++ b/contrib/mkinitcpio/sbctl
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+echo "Signing EFI binaries..."
+/usr/bin/sbctl sign-all -g
--- a/contrib/pacman/ZZ-sbctl.hook
+++ b/contrib/pacman/ZZ-sbctl.hook
@@ -7,10 +7,7 @@ Target = boot/*
 Target = efi/*
 Target = usr/lib/modules/*/vmlinuz
 Target = usr/lib/modules/*/extramodules/*
-Target = usr/lib/initcpio/*
 Target = usr/lib/**/efi/*.efi*
-Target = usr/lib/firmware/*
-Target = usr/src/*/dkms.conf
 
 [Action]
 Description = Signing EFI binaries...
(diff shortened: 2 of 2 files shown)


Foxboron (approved):

agowa:
How about also adding an initcpio hook? That way it it almost guaranteed to call sbctl every time when the kernel gets rebuilt. Regardless what triggered it (or if the user itself triggered it and forgot that they have to sign the kernel too, *not me*)

Also see #218

Foxboron:
At that point I'd rather just deprecate the pacman hook and replace with kernel-install or mkinitcpio hooks.

agowa:
It doesn't hurt to have it. I kinda like have the "double protection" of two independent hooks making sure that the image is really signed. But yea, the pacman one is then basically completely useless...

Joseph-DiGiovanni:
I agree the mkinitcpio post hook is the way to go. However, I think the best solution is to just keep the sign-all command from the pacman hook rather than the hook making assumptions of what files should be signed.

This should close #263 and #218.

Foxboron:
Actually, I realized this would be a bad idea, maybe.

We do sign more then just the initrd. Namely things like `sd-boot` and `fwudpmgr`. And I'm not sure if signing those as part of a post-hook in mkinitcpio makes sense?

Joseph-DiGiovanni:
Good point. So we should keep both. For the sake of not breaking anything we should keep it as is with the exception of the usr/lib/initcpio/* target and my other additions. It will attempt to sign twice most of the time but this is preferable to not signing at all.

agowa:
> And I'm not sure if signing those as part of a post-hook in mkinitcpio makes sense?

It doesn't hurt and sbctl noops already if these files are already signed.

I'd say still stick with it is better to have a potentially useless pacman hook in addition to the initcpio one than risking a non bootable system because of unforeseen circumstances....

contrib/mkinitcpio/sbctl
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
Foxboron, line 1:
```suggestion
#!/usr/bin/bash
```

Foxboron (changes requested):
lgtm

Foxboron:
Thanks!

Foxboron (approved):

PR #288, opened 2024-02-22
ci.yaml: use github container registry 

pheiduck:
See: https://github.com/Foxboron/sbctl/actions/runs/8008805592/job/21876043359#step:3:7

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
          - run: GOBIN=/usr/bin make lint
    void:
       runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Foxboron:
I suspect getting `sudo` without syncing the repos is going to be a bit hard :)

pheiduck:
> I suspect getting `sudo` without syncing the repos is going to be a bit hard :)

There is also a certificate error
Danger option: `SSL_NO_VERIFY_PEER=true`

pheiduck:
The ghcr was updated 2month ago.
https://github.com/void-linux/void-containers/pkgs/container/void-musl

PR #298, opened 2024-04-23
[WIP] Introduce prepare-integration make target

jimmykarily:
to prepare the VM filesystem for the tests.
This target is suppused to be run like so:

make prepare-integration integration

but currently the filesystem it creates is not bootable for some reason.

@Foxboron looking at the code it seems that there is some way to run the integration test but it's relying on things outside git. I tried to fill in the missing parts by wrapping the make_image.sh script in a docker command that would make it possible to run it locally.

I remember you mentioned u-root being used in `go-uefi` so I wonder if I should ditch this and try to use `u-root` to build an initramfs that has sbctl in it. What do you think? I'm also fine dropping this completely until you find time to work on it yourself. I only mean to help, no pressure :).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,10 @@ lint:
 test:
 	go test -v ./...
 
+.PHONY: prepare-integration
+prepare-integration:
+	docker run --privileged -v ${PWD}/tests/:/workdir -w /workdir archlinux ./make_image.sh
+
 .PHONY: integration
 integration:
 	go test -v tests/integration_test.go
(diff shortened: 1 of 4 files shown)


Foxboron:
Thanks for taking stabs at the project :)

> I remember you mentioned u-root being used in go-uefi so I wonder if I should ditch this and try to use u-root to build an initramfs that has sbctl in it. What do you think? I'm also fine dropping this completely until you find time to work on it yourself. I only mean to help, no pressure :).

I think moving the current tests from the old `vmtest` library to the u-root library would make all of this simpler.

jimmykarily:
> Thanks for taking stabs at the project :)
> 
> > I remember you mentioned u-root being used in go-uefi so I wonder if I should ditch this and try to use u-root to build an initramfs that has sbctl in it. What do you think? I'm also fine dropping this completely until you find time to work on it yourself. I only mean to help, no pressure :).
> 
> I think moving the current tests from the old `vmtest` library to the u-root library like I've done in `go-uefi` would make all of this simpler.

Mind me giving it a try?

Foxboron:
> Mind me giving it a try?

Sure! Feel free :)

jimmykarily:
Closing this PR to try a different approach as described above.

Issue #311, opened 2024-05-12
pacstrap fails due to mkinitcpio post hook


punoko:
Hello,

pacstrap exits with the following error when invoked with sbctl in the package list, as it is apparently trying to sign before keys have even been created :

```
# pacstrap -cKM ./sbctl base linux sbctl
...
==> Creating zstd-compressed initcpio image: '/boot/initramfs-linux-fallback.img'
  -> Early uncompressed CPIO image generation successful
==> Initcpio image generation successful
==> Running post hooks
  -> Running post hook: [sbctl]
Signing /boot/vmlinuz-linux
couldn't access /usr/share/secureboot/keys/db/db.pem: no such file or directory
==> ERROR: '/usr/lib/initcpio/post/sbctl' failed with exit code 1
error: command failed to execute correctly
```

<details>
  <summary>Full log</summary>

```
[root@60753b2172bf /]# truncate -s 2G sbctl.img

[root@60753b2172bf /]# losetup -fP --show sbctl.img
/dev/loop0

[root@60753b2172bf /]# mkfs.btrfs /dev/loop0
btrfs-progs v6.8.1
See https://btrfs.readthedocs.io for more information.

Performing full device TRIM /dev/loop0 (2.00GiB) ...
NOTE: several default settings have changed in version 5.15, please make sure
      this does not affect your deployments:
      - DUP for metadata (-m dup)
      - enabled no-holes (-O no-holes)
      - enabled free-space-tree (-R free-space-tree)

Label:              (null)
UUID:               7ca9a7f5-5c0a-49c1-a38d-10ceb86060fc
Node size:          16384
Sector size:        4096	(CPU page size: 4096)
Filesystem size:    2.00GiB
Block group profiles:
  Data:             single            8.00MiB
  Metadata:         DUP             102.38MiB
  System:           DUP               8.00MiB
SSD detected:       yes
Zoned device:       no
Features:           extref, skinny-metadata, no-holes, free-space-tree
Checksum:           crc32c
Number of devices:  1
Devices:
   ID        SIZE  PATH
    1     2.00GiB  /dev/loop0

[root@60753b2172bf /]# mkdir sbctl

[root@60753b2172bf /]# mount -o noatime,compress /dev/loop0 ./sbctl

[root@60753b2172bf /]# pacstrap -cKM ./sbctl base linux sbctl
==> Creating install root at ./sbctl
gpg: /./sbctl/etc/pacman.d/gnupg/trustdb.gpg: trustdb created
[...]

PR #312, opened 2024-05-12
Update the mkinitcpio post hook to return 0 when no keys are found

punoko:
Fixes https://github.com/Foxboron/sbctl/issues/311


Foxboron:
Thanks!

PR #342, opened 2024-07-31
sbctl: Fix human readable output being printed when using `--json`

chenxiaolong:
`PersistentPreRun` was being overwritten later in `main()`, causing `logging.PrintOff()` to never be called.

EDIT: Updated with @Foxboron's suggestion below.


Foxboron:
We should be setting a global `on` variable in the logging path to toggle printing and no printing.

It's unclear to me why this doesn't work.

https://github.com/Foxboron/sbctl/blob/master/logging/logging.go#L50

Foxboron:
Ah, I found the issue. I overwrite `PersistentPreRun` in the main function. If you could move the code here https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/main.go#L68 to here https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/main.go#L136 that would fix the issue :)

chenxiaolong:
Ah, I totally missed that the logging system itself was checking the flag. Thanks! I'll update the PR with your suggestion.

chenxiaolong:
Done!

Foxboron:
Thanks!

PR #350, opened 2024-08-05
sbctl.8: fix typo in the man page

acuteenvy:
The same typo is in the release notes for 0.15:
> Note that only RSA 2048 is supported by most TPMs while the default `file` type
defaults to RSA 409***8***.


Foxboron:
Ufff, that is embarrassing. Thanks for the fix!

PR #352, opened 2024-08-05
sbctl.hook: quieten output to reduce terminal spam

hobbitalastair:
For systems dual-booting Windows, there are a large number of signed EFI files. Avoid outputting "File has already been signed" for each of them.

### DIFF
--- a/contrib/pacman/ZZ-sbctl.hook
+++ b/contrib/pacman/ZZ-sbctl.hook
@@ -12,4 +12,4 @@ Target = usr/lib/**/efi/*.efi*
 [Action]
 Description = Signing EFI binaries...
 When = PostTransaction
-Exec = /usr/bin/sbctl sign-all -g
+Exec = /usr/bin/sbctl sign-all -g --quiet


hobbitalastair:
I like the tool, by the way! Much easier than the last time I mucked around with secure boot.

My system wouldn't boot after the last update, which appears to be because I'd missed the "failed creating bundle" message in the pacman output the last time I updated.
For the sake of making it easier to spot, it would be nice if there wasn't a wall of text every time I updated. If I read the code right, `--quiet` should only print errors, which is perhaps less reassuring but more useful.

This may be user error because I've signed all the `efi.mui` translation files from Microsoft - all 94 of them. It'd be a lot more manageable without that, only 23 lines or so of output.

Foxboron:
>My system wouldn't boot after the last update, which appears to be because I'd missed the "failed creating bundle" message in the pacman output the last time I updated.

Should be fixed with the recent release. I didn't buble up the errors in `sign-all` so any failing signatures would no exit with a non-zero code.

https://github.com/Foxboron/sbctl/commit/ff13e7c2cb00200a5c973c5b609266de8dc6acf8

> This may be user error because I've signed all the efi.mui translation files from Microsoft - all 94 of them. It'd be a lot more manageable without that, only 23 lines or so of output.

Why do you sign these?

hobbitalastair:
> > This may be user error because I've signed all the efi.mui translation files from Microsoft - all 94 of them. It'd be a lot more manageable without that, only 23 lines or so of output.
> 
> Why do you sign these?

Because I figured they needed to be signed (surely MS wouldn't load them otherwise?).

It turns out that:
- they are signed with the microsoft keys already (so I assume they do need to be signed)
- I have kept the microsoft keys
So I have happily removed them from sbctl's tender care!

I still think that sbctl should be less verbose when run as a pacman hook. And someone without the microsoft keys enrolled would have to work a bit harder here (although maybe they wouldn't be dual booting). But my particular use case is probably solved; feel free to close this PR if you don't think it's worth fixing.

PR #363, opened 2024-08-15
Allow exporting keys without efivars present and update CI staticcheck

Cornelicorn:
Closes https://github.com/Foxboron/sbctl/issues/364


cmd/sbctl/enroll-keys.go
```diff
@@ -255,7 +255,7 @@ func KeySync(state *config.State, oems []string) error {
 func RunEnrollKeys(state *config.State) error {
 	// SetupMode is not necessarily required on a partial enrollment
 	ok, err := state.Efivarfs.GetSetupMode()
-	if err != nil {
+	if err != nil && enrollKeysCmdOptions.Export.Value == "" {
```
Cornelicorn, line 258: This doesn't look very nice, but the alternative would be to nest the next few lines inside `if enrollKeysCmdOptions.Export.Value == ""  {` which is also not very nice, happy to reformat this if you have a better suggestion.

Foxboron: This needs a comment, else I'll totally not get the context later.

Foxboron:
Generally it looks fine, but what is the usecase?

Cornelicorn:
> Generally it looks fine, but what is the usecase?

Just added the issue, maybe should do that the other way around next time :D

Foxboron:
Ah, cool. This makes sense.

But then I suspect it would be better to entierly decouple the export stuff from the enroll functions. Currently they are quite entangled but there isn't a reason why everything is refactored into a `RunExportKeys` and then we just have an early if-check for a non-empty `Export`.

We don't really care about whether or not enrollment fails to export the keys, and exporting them could probably be done before enrollment.

Don't have to do this for this PR though, but it would be trivial to write and also test for in the current setup.

Foxboron (approved):

Cornelicorn:
> Ah, cool. This makes sense.
> 
> But then I suspect it would be better to entierly decouple the export stuff from the enroll functions. Currently they are quite entangled but there isn't a reason why everything is refactored into a `RunExportKeys` and then we just have an early if-check for a non-empty `Export`.
> 
> We don't really care about whether or not enrollment fails to export the keys, and exporting them could probably be done before enrollment.
> 
> Don't have to do this for this PR though, but it would be trivial to write and also test for in the current setup.

Yes, enrolling and exporting should probably be split up.
I won't be able to get started with this in the next 4 weeks sadly, but the error is relatively easy to reproduce with mkosi:

mkosi.conf
```
[Distribution]
Distribution=arch

[Content]
Bootable=yes
Packages=
    linux
    systemd
```

mkosi.postinst.chroot
```
#!/bin/bash

sbctl create-keys
mkdir -p /efi/loader/keys/auto
cd /efi/loader/keys/auto
sbctl enroll-keys --disable-landlock --custom --yes-this-might-brick-my-machine --export auth
sbctl enroll-keys --disable-landlock --custom --yes-this-might-brick-my-machine --export esl
```

Then run
```
mkdir -p mkosi.extra/usr/bin
cp /my/sbctl/binary mkosi.extra/usr/bin/sbctl
mkosi
```

Foxboron:
> I won't be able to get started with this in the next 4 weeks sadly

We'll see if I have time for it if I get bored for an hour. Should not be very hard.

I'll accept this PR regardless though. Gives me a little bit of motivation to do it :)

Foxboron:
Thanks!

PR #377, opened 2024-09-29
Read output file path from db if output not specified

bkrochta:
The mkinitcpio hook runs `sbctl sign "$IMAGE_FILE"` and has no way to know the output file path for the corresponding image file. This results in signing in place which may break booting for those who with different locations for signed and unsigned UKIs.

The sign command should check if the given file is in the database and use the saved output file path.

### DIFF
--- a/cmd/sbctl/sign.go
+++ b/cmd/sbctl/sign.go
@@ -38,6 +38,21 @@ var signCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		// Get output path from database for file if output not specified
+		if output == "" {
+			files, err := sbctl.ReadFileDatabase(state.Fs, state.Config.FilesDb)
+			if err != nil {
+				return err
+			}
+			for _, entry := range files {
+				if entry.File == file {
+					logging.Print("Using output file from database\n")
+					output = entry.OutputFile
+					break
+				}
+			}
+		}
+
 		if output == "" {
 			output = file
 			rules = append(rules, lsm.TruncFile(file).IgnoreIfMissing())
--- a/cmd/sbctl/sign.go
+++ b/cmd/sbctl/sign.go
@@ -46,7 +46,6 @@ var signCmd = &cobra.Command{
 			}
 			for _, entry := range files {
 				if entry.File == file {
-					logging.Print("Using output file from database\n")
 					output = entry.OutputFile
 					break
 				}


Foxboron:
I don't quite understand this patch. In which case would `$IMAGE_FILE` point at a file which should be placed somewhere else? Using `sbctl` this way to throw a file into the correct location this way seems wrong?

bkrochta:
In my circumstance, I am running Arch, and I have unsigned unified kernel images built by `mkinitcpio` in `/boot/unsigned-images/` on my encrypted root partition where I then want to have them signed and placed in the unencrypted EFI partition.

When first setting up `sbctl`, I signed all my images with `sbctl sign -s /boot/unsigned-images/... -o /efi/EFI/...` so that they are saved in the `sbctl` database. However, this database is only read and used when running `sbctl sign-all` and not `sbctl sign`. 

When the UKIs are triggered to be rebuilt by `mkinitcpio`, the hook from `sbctl` runs and signs each image as it is built. However, it ignores the output location I had saved and signs the image in-place. This results in the images in the EFI not being updated which may cause the computer to fail to boot if the kernel doesn't match what is installed on the computer.

There is a pacman hook that runs `sbctl sign-all`, but this is broken in the current release saying the "image is already signed", but after pulling the most recent changes since `0.15.4`, it has been fixed and it does move the re-sign the files and move them to the EFI. However, the `mkinitcpio` hook is still signing images in place when the database specifies an out location. If I manually run `mkinitcpio -p ...`, I'd still have to move the file to my EFI.

I know this reasoning Arch specific, but I do believe if someone had initially run something like `sbctl sign -s /boot/unsigned-images/... -o /efi/EFI/...` to save the output location, that subsequent `sbctl sign`s should respect that saved location. If they want to manually put it somewhere else or sign in place just this one time, then they can use the `-o` option.

cmd/sbctl/sign.go
```diff
@@ -38,6 +38,21 @@ var signCmd = &cobra.Command{
...
 		}
+		// Get output path from database for file if output not specified
+		if output == "" {
+			files, err := sbctl.ReadFileDatabase(state.Fs, state.Config.FilesDb)
+			if err != nil {
+				return err
+			}
+			for _, entry := range files {
+				if entry.File == file {
+					logging.Print("Using output file from database\n")
```
Foxboron, line 49: Remove this line, or move it as a debug print.

cmd/sbctl/sign.go
```diff
@@ -38,6 +38,21 @@ var signCmd = &cobra.Command{
...
 			return err
 		}
+		// Get output path from database for file if output not specified
+		if output == "" {
+			files, err := sbctl.ReadFileDatabase(state.Fs, state.Config.FilesDb)
+			if err != nil {
+				return err
+			}
+			for _, entry := range files {
+				if entry.File == file {
```
Foxboron, line 48: Can probably use the slice package, but this is fine.

PR #392, opened 2024-10-19
build.yml: download artifacts  using a wildcard

pheiduck:
See: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md#supported-characters


Foxboron (approved):

Foxboron:
If you want another free PR then the same issue probably exists in the age workflow :)

https://github.com/FiloSottile/age/blob/main/.github/workflows/build.yml#L80

PR #393, opened 2024-10-20
Fix input/output file comparison when signing

bkrochta:
`Stat` doesn't return `ErrExist` error, so this line is always false. We also don't need any stats for the output file, so we can just try to open the file, and continue the comparison with the input file if there is no error opening it.


Foxboron:
Thanks! I plan on removing all `stat` calls and just use `open` directly at some point.

PR #402, opened 2024-11-21
Verified JSON support

cybik:
Implement rather simple but functional JSON output mode for verify.

### DIFF
--- a/cmd/sbctl/verify.go
+++ b/cmd/sbctl/verify.go
@@ -16,6 +16,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type VerifiedFile struct {
+	FileName           string         `json:"file_name"`
+	FileIsSigned       bool           `json:"is_signed"`
+}
+
 var (
 	ErrInvalidHeader = errors.New("invalid pe header")
 	verifyCmd        = &cobra.Command{
@@ -23,6 +28,7 @@ var (
 		Short: "Find and check if files in the ESP are signed or not",
 		RunE:  RunVerify,
 	}
+	verifiedFiles    = make([]VerifiedFile, 0)
 )
 
 func VerifyOneFile(state *config.State, f string) error {
@@ -52,11 +58,16 @@ func VerifyOneFile(state *config.State, f string) error {
 	if err != nil {
 		return err
 	}
+
+	file_entry := &VerifiedFile{FileName: f, FileIsSigned: false}
 	if ok {
 		logging.Ok("%s is signed", f)
+		file_entry.FileIsSigned = true
 	} else {
 		logging.NotOk("%s is not signed", f)
 	}
+	verifiedFiles = append(verifiedFiles, *file_entry)
+
 	return nil
 }
 
@@ -88,6 +99,11 @@ func RunVerify(cmd *cobra.Command, args []string) error {
 					logging.Error(fmt.Errorf("%s is not a valid EFI binary", file))
 					return nil
 				}
+				return err
+			}
+		}
+		if cmdOptions.JsonOutput {
+			if err := JsonOut(verifiedFiles); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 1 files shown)


cmd/sbctl/verify.go
```diff
@@ -16,13 +16,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type VerifiedFile struct {
+	FileName           string         `json:"file_name"`
+	FileIsSigned       bool           `json:"is_signed"`
```
Foxboron, line 21:
```suggestion
	IsSigned       bool           `json:"is_signed"`
```

cmd/sbctl/verify.go
```diff
@@ -91,6 +102,11 @@ func RunVerify(cmd *cobra.Command, args []string) error {
 				return err
 			}
 		}
+		if cmdOptions.JsonOutput {
+			if err := JsonOut(verifiedFiles); err != nil {
+				return err
+			}
```
Foxboron, line 108:
```suggestion
			return JsonOut(verifiedFiles);
```

cmd/sbctl/verify.go
```diff
@@ -16,13 +16,19 @@ import (
...
+}
+
 var (
 	ErrInvalidHeader = errors.New("invalid pe header")
 	verifyCmd        = &cobra.Command{
 		Use:   "verify",
 		Short: "Find and check if files in the ESP are signed or not",
 		RunE:  RunVerify,
 	}
+	verifiedFiles    = make([]VerifiedFile, 0)
```
Foxboron, line 31:
```suggestion
	verifiedFiles    []*VerifiedFile
```

I think?

cmd/sbctl/verify.go
```diff
@@ -52,11 +58,16 @@ func VerifyOneFile(state *config.State, f string) error {
 	if err != nil {
 		return err
 	}
+
+	file_entry := &VerifiedFile{FileName: f, FileIsSigned: false}
```
Foxboron, line 62:
```suggestion
	fileentry := &VerifiedFile{FileName: f, IsSigned: false}
```

Prefer CamelCase in Go :) Note that `IsSigned` is initialized as "false", so you don't have to write it at this point.

cmd/sbctl/verify.go
```diff
@@ -52,11 +58,16 @@ func VerifyOneFile(state *config.State, f string) error {
...
 	if ok {
 		logging.Ok("%s is signed", f)
+		file_entry.FileIsSigned = true
 	} else {
 		logging.NotOk("%s is not signed", f)
 	}
+	verifiedFiles = append(verifiedFiles, *file_entry)
```
Foxboron, line 69:
```suggestion
	verifiedFiles = append(verifiedFiles, fileentry)
```

Foxboron:
Thanks! Generally looks good to me. If you want to try and write a test for this please do :)

Pointers from similar code:

https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/status_test.go#L172

https://github.com/Foxboron/sbctl/blob/master/cmd/sbctl/setup_test.go#L27

But if this is not something you are comfortable with then just writing the `--json` support is fine as well.

cybik:
I'll look at writing the test. Honestly this was my first PR in Go ever :P

I was also considering changing the state to a byte or a string because there's 3 states in verify - signed, unsigned and missing (but you knew that lol) - the latter of which I missed initially.

Foxboron:
> I'll look at writing the test. Honestly this was my first PR in Go ever :P

It's not a very complicated language so easy to get started!

> I was also considering changing the state to a byte or otherwise, because there's 3 states in verify - signed, unsigned and missing (but you knew that lol) - the latter of which I missed initially.

I think just checking if there is a valid signature or not is fine really. The alternative would be to have the field as a Go enum.

But that is probably a bit more involved to get correctly.

https://gist.github.com/lummie/7f5c237a17853c031a57277371528e87

cybik:
I pushed an update with the field as an int. To be honest, if the json mode is being used, chances are the data is going to be processed by other tools, so just adding documentation about what the outputs mean is probably enough?

cybik:
I'll be frank, writing up a test case for this might be slightly more involved since there's no existing code for `verify` that shims out the necessary stuff to test the existing `verify` code.

So I'll convert this PR back to "ready".

Foxboron (approved):

Foxboron:
Thanks!

PR #409, opened 2024-12-23
initial yubikey piv support

tomis007:
Hey! Super cool project. I wanted to add yubikey support with its `piv` feature. Here's my initial PR for that. It's a work in progress -- please do not merge the code is still hacky, but I made the PR to get your take on my implementation so far. A few notes:

* the `go-piv/piv-go` project has a bug where it doesn't actually export a few of the key types because they have lowercase names. I submitted https://github.com/go-piv/piv-go/pull/164 to fix that
* I'm storing `priv` in the config passed to the yubikey type but I am not sure if that's needed or if it's better to just prompt every time instead
* I put one private key in the 9C slot (https://docs.yubico.com/yesdk/users-manual/application-piv/slots.html) and used that for all 3 of the required keys that are generated for secure boot. Is there a better way to do this? Maybe 3 different ones in the other slots? My current implementation works as it requires a pin for every signature with the key https://developers.yubico.com/PIV/Introduction/Certificate_slots.html
* I sort of had to hack it together because it doesn't look like the config file is finished yet. I could fix that up too to make it easier to use the yubikey piv signing 
* I'm not storing the private key on disk, instead storing the public key and using that to request the private key from the yubikey when needed. I think that makes the most sense

Thanks!

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -299,18 +299,16 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
 }
 
 func GetBackendType(b []byte) (BackendType, error) {
-	block, _ := pem.Decode(b)
-	if block == nil {
-		//the yubikey data is json
+	if json.Valid(b) {
 		var yubiData YubikeyData
 		err := json.Unmarshal(b, &yubiData)
 		if err != nil {
 			logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
 			return "", err
 		}
-		fmt.Printf("unmarshalled!: %s\n", b)
 		return YubikeyBackend, nil
 	} else {
+		block, _ := pem.Decode(b)
 		// TODO: Add TSS2 keys
 		switch block.Type {
 		case "PRIVATE KEY":
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -81,8 +81,6 @@ func (k *KeyHierarchy) GetKeyBackend(e efivar.Efivar) KeyBackend {
 		return k.KEK
 	case efivar.Db:
 		return k.Db
-		// case efivar.Dbx:
-		// 	return k.Dbx
 	default:
 		panic("invalid key hierarchy")
 	}
@@ -200,6 +198,8 @@ func createKey(state *config.State, backend string, hier hierarchy.Hierarchy, de
 		return NewFileKey(hier, desc)
 	case "tpm":
 		return NewTPMKey(state.TPM, desc)
+	case "yubikey":
+		return NewYubikeyKey(state.YubikeySigKeys, desc)
 	default:
 		return NewFileKey(hier, desc)
 	}
@@ -255,6 +255,8 @@ func readKey(state *config.State, keydir string, kc *config.KeyConfig, hier hier
 		return FileKeyFromBytes(keyb, pemb)
 	case TPMBackend:
 		return TPMKeyFromBytes(state.TPM, keyb, pemb)
+	case YubikeyBackend:
+		return YubikeyFromBytes(state.YubikeySigKeys, keyb, pemb)
 	default:
 		return nil, fmt.Errorf("unknown key")
 	}
@@ -302,6 +304,8 @@ func GetBackendType(b []byte) (BackendType, error) {
 		return FileBackend, nil
 	case "TSS2 PRIVATE KEY":
 		return TPMBackend, nil
+	case "PUBLIC KEY":
+		return YubikeyBackend, nil
 	default:
 		return "", fmt.Errorf("unknown file type: %s", block.Type)
 	}
@@ -322,6 +326,8 @@ func InitBackendFromKeys(state *config.State, priv, pem []byte, hier hierarchy.H
 		return FileKeyFromBytes(priv, pem)
 	case "tpm":
 		return TPMKeyFromBytes(state.TPM, priv, pem)
+	case "yubikey":
+		return YubikeyFromBytes(state.YubikeySigKeys, priv, pem)
 	default:
 		return nil, fmt.Errorf("unknown key backend: %s", t)
 	}
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -3,9 +3,11 @@ package backend
 import (
 	"crypto"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"github.com/foxboron/sbctl/logging"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


backend/yubikey.go
```diff
@@ -0,0 +1,206 @@
...
+	"strings"
+	"time"
+
+	"github.com/tomis007/piv-go/v2/piv"
+)
+
+type Yubikey struct {
+	keytype BackendType
+	cert    *x509.Certificate
+	privkey *piv.KeyRSA
```
Foxboron, line 24:
```suggestion
	privkey crypto.Signer
```

Which is why I suspect the upstream author disagrees. This also generalizes the support for RSA and NIST keys.

backend/yubikey.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil, fmt.Errorf("yubikey key not found")
+		}
+		conf.YK = yk
+
+		keyInfo, err := yk.KeyInfo(piv.SlotSignature)
+		if err != nil {
+			return nil, err
+		}
+		if keyInfo.PublicKey != nil {
+			if keyInfo.Algorithm == piv.AlgorithmRSA2048 {
```
Foxboron, line 56: Use a switch/case statement

backend/yubikey.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+			logging.Println("Creating key... please press Yubikey")
+			pub, err = yk.GenerateKey(piv.DefaultManagementKey, piv.SlotSignature, key)
+			if err != nil {
+				return nil, err
+			}
+
+		}
+
+		// TODO prompt for PIN when it's not default
```
Foxboron, line 77: This can probably be left as a TODO until I figure out the PIN/password support.

backend/yubikey.go
Foxboron: I write while recalling that Secure Boot only support RSA keys :)

backend/yubikey.go
Foxboron: N/A, we only want RSA keys!

Foxboron:
Thanks for working on this!

> the go-piv/piv-go project has a bug where it doesn't actually export a few of the key types because they have lowercase names. I submitted https://github.com/go-piv/piv-go/pull/164 to fix that

I agree with upstream and you should just use the `crypto.Signer` interface.

> I'm storing priv in the config passed to the yubikey type but I am not sure if that's needed or if it's better to just prompt every time instead

Prompt everytime I reckon. We want to be able to have a different backend pr type so it makes sense to not pass this around.


> I put one private key in the 9C slot (https://docs.yubico.com/yesdk/users-manual/application-piv/slots.html) and used that for all 3 of the required keys that are generated for secure boot. Is there a better way to do this? Maybe 3 different ones in the other slots? My current implementation works as it requires a pin for every signature with the key https://developers.yubico.com/PIV/Introduction/Certificate_slots.html

Preferably one key pr key on the yubikey. This reflects the proper key hierarchy and allows us flexibility.

> I sort of had to hack it together because it doesn't look like the config file is finished yet. I could fix that up too to make it easier to use the yubikey piv signing

Preferably work on the config parts of it, and if anything is too difficult or problematic please just tell me and I'll write up the code you need.

> I'm not storing the private key on disk, instead storing the public key and using that to request the private key from the yubikey when needed. I think that makes the most sense

I wonder if we should store some extra metadata around the key. So we can properly look it up when needed. What happens if we have multiple key?

backend/backend.go
```diff
@@ -302,6 +304,8 @@ func GetBackendType(b []byte) (BackendType, error) {
 		return FileBackend, nil
 	case "TSS2 PRIVATE KEY":
 		return TPMBackend, nil
+	case "PUBLIC KEY":
+		return YubikeyBackend, nil
```
Foxboron, line 308: We should probably do something more identifiable here, or restructure it so we can figure out a better way to detect keys.

backend/yubikey.go
tomis007: Ah yes I see the issue, I was unable to figure out type casting the `priv` to crypto.Signer. Updated to fix that and to not use my fork

backend/yubikey.go
tomis007: I think here I'll want it to prompt for the Yubikey PIV pin when it's not `123456`. Were you talking about a different PIN?

tomis007:
Thanks for the review! I'll keep working on it. I am thinking of using a .toml config file instead of .yaml, do you have any strong preference? I personally don't really like using .yaml as a config format because of how complicated it can be

backend/yubikey.go
Foxboron: I'm thinking generally how to deal with password prompts in sbctl. It's not a thing yet and that needs to be planned a bit.

Foxboron:
I'd prefer if we stick with `sbctl.conf`. there isn't a need to invent something new only for yubikey. If you need to serialize some state you can just use json

tomis007:
> I'd prefer if we stick with `sbctl.conf`. there isn't a need to invent something new only for yubikey. If you need to serialize some state you can just use json

Will do! I

tomis007:
I've been testing out my PR branch with the following `sbctl.conf` and the `sbctl create-keys --type yubikey` command. I noticed there's no one option in the `.conf` to set a directory for everything so I added that as `datadir`. Also the config parsing code currently does not attempt to make directories. I think the PR is ready for another review / to see what its missing. I also need to update the dependencies in the readme for the yubikey packages. I updated it to store `.json` data for the yubikey information. Let me know what other changes you think it needs

```yaml
datadir: /var/lib/sbctl/yubikeys

db_additions:
  - microsoft
files:
  - path: /boot/vmlinuz-linux-lts
  - path: /usr/lib/fwupd/efi/fwupdx64.efi
    output: /usr/lib/fwupd/efi/fwupdx64.efi.signed
keys:
  pk:
    type: yubikey
  kek:
    type: yubikey
  db:
    type: yubikey
```

backend/yubikey.go
tomis007: I ended up using `https://github.com/manifoldco/promptui` in my yubikey backend file, which works well and fits the general aesthetic / workflow of `sbctl`

Foxboron:
I'm busy at congress and stuff until new years. So I can't promise I'll get to this until over new years :)

tomis007:
no worries! have a good new years!

tomis007:
I did a `pacman -Syu` today and noticed something with the post hooks for mkinitcpio and pacman. According to https://bbs.archlinux.org/viewtopic.php?id=255944, pacman explicitly does not allow prompting for user input during a post script and all the scripts aren't run in ttys. Unsure exactly what the best way to ask for the yubikey pin in this case is as no matter what `stdin` is set to `^D` / EOF. It might be easiest to add a message to sbctl yubikey signing that if it's not running in a `tty` to print without doing anything and output a message telling you to rerun the command yourself, let me know what you think

peto59:
Might I suggest `systemd-ask-password`? It does have `--no-tty` flag and with `--timeout=0` it will wait indefinitely. I use it to query passwords during boot which has similar constraints.

tomis007:
Oh that’s cool, it looks like it might work. I will test it out, thanks!

tomis007:
I fixed the non tty issue with `systemd-ask-password`

tomis007:
I think it's ready for another review, I have it working locally on my machine for signing my uki and with the hooks, but further testing from someone else would be good

Foxboron:
Would it be possible for you to work a bit more on the commit history of this PR? Having 13 commits with non-descriptive names makes it harder to review properly.

tomis007:
Sure, let me know if that helps. It's probably best to focus on the final commit most. I didn't bother squashing the commits for the pr

Foxboron:
Fwiw, I intend to review this! But please note I'm a bit busy the next two weeks leading up to FOSDEM :)

tomis007:
Sounds good, thanks!

tomis007:
I may have messed up renaming the commits from my second computer it seems like the number has doubled, feel free to just ask me for further clarification once you review if that's helpful

Foxboron:
It looks like you have rebased ontop of your... old branch. This is getting a bit unwieldy to review. 

I'd suggest unstaging all commits, and commit every piece by piece so every commit is an atomic change.

PR #412, opened 2025-01-12
fix: print an error on setup without args

acuteenvy:
`sbctl setup` exits with status code 0 and doesn't print anything. Because one argument is required (`--migrate`, `--print-config`, `--print-state` or `--setup`), `sbctl setup` without other options should be an error.


Foxboron (approved):

Foxboron:
Thanks!

PR #413, opened 2025-01-20
Yubikey piv -- take 2 

tomis007:
Closing #409 and replacing it with this one with a cleaner git commit history

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -301,8 +301,7 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
 func GetBackendType(b []byte) (BackendType, error) {
 	if json.Valid(b) {
 		var yubiData YubikeyData
-		err := json.Unmarshal(b, &yubiData)
-		if err != nil {
+		if err := json.Unmarshal(b, &yubiData); err != nil {
 			logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
 			return "", err
 		}
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -299,18 +299,16 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
 }
 
 func GetBackendType(b []byte) (BackendType, error) {
-	block, _ := pem.Decode(b)
-	if block == nil {
-		//the yubikey data is json
+	if json.Valid(b) {
 		var yubiData YubikeyData
 		err := json.Unmarshal(b, &yubiData)
 		if err != nil {
 			logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
 			return "", err
 		}
-		fmt.Printf("unmarshalled!: %s\n", b)
 		return YubikeyBackend, nil
 	} else {
+		block, _ := pem.Decode(b)
 		// TODO: Add TSS2 keys
 		switch block.Type {
 		case "PRIVATE KEY":
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -81,8 +81,6 @@ func (k *KeyHierarchy) GetKeyBackend(e efivar.Efivar) KeyBackend {
 		return k.KEK
 	case efivar.Db:
 		return k.Db
-		// case efivar.Dbx:
-		// 	return k.Dbx
 	default:
 		panic("invalid key hierarchy")
 	}
@@ -200,6 +198,8 @@ func createKey(state *config.State, backend string, hier hierarchy.Hierarchy, de
 		return NewFileKey(hier, desc)
 	case "tpm":
 		return NewTPMKey(state.TPM, desc)
+	case "yubikey":
+		return NewYubikeyKey(state.YubikeySigKeys, desc)
 	default:
 		return NewFileKey(hier, desc)
 	}
@@ -255,6 +255,8 @@ func readKey(state *config.State, keydir string, kc *config.KeyConfig, hier hier
 		return FileKeyFromBytes(keyb, pemb)
 	case TPMBackend:
 		return TPMKeyFromBytes(state.TPM, keyb, pemb)
+	case YubikeyBackend:
+		return YubikeyFromBytes(state.YubikeySigKeys, keyb, pemb)
 	default:
 		return nil, fmt.Errorf("unknown key")
 	}
@@ -302,6 +304,8 @@ func GetBackendType(b []byte) (BackendType, error) {
 		return FileBackend, nil
 	case "TSS2 PRIVATE KEY":
 		return TPMBackend, nil
+	case "PUBLIC KEY":
+		return YubikeyBackend, nil
 	default:
 		return "", fmt.Errorf("unknown file type: %s", block.Type)
 	}
@@ -322,6 +326,8 @@ func InitBackendFromKeys(state *config.State, priv, pem []byte, hier hierarchy.H
 		return FileKeyFromBytes(priv, pem)
 	case "tpm":
 		return TPMKeyFromBytes(state.TPM, priv, pem)
+	case "yubikey":
+		return YubikeyFromBytes(state.YubikeySigKeys, priv, pem)
 	default:
 		return nil, fmt.Errorf("unknown key backend: %s", t)
 	}
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -3,9 +3,11 @@ package backend
 import (
 	"crypto"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"github.com/foxboron/sbctl/logging"
 	"io"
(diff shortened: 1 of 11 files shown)


tomis007:
@Foxboron I cleaned up the git commit messages and just started fresh with a new branch, I think it's now much cleaner

Foxboron:
The commit messages are better, thank you! But there are several WIP commits in the middle of this PR that doesn't serve any purpose to the final code.

tomis007:
Yes that's accurate. What would be most helpful for review? The final state is really the state to review given that the intermediate commits have bugs/was under development. I could squash all the commits if you want (I assume you will squash the PR before merging regardless?)

Foxboron:
Removing the intermediate debugging commits would be a great start. Then ensure each commit is preferably doing one thing and atomic so they can be bisected for issues at a later point.

> I assume you will squash the PR before merging regardless?

Not really. I prefer smaller atomic commits introducing one and one change instead of having one commit with several changes. Makes for easier debugging, but I'm not super strict on it.

tomis007:
That makes sense, I'll restructure the PR/commits to be a few atomic changes that can all be merged rather than one large squashed change. Probably within the next week or so

Foxboron:
Thanks :)!

backend/backend.go
```diff
@@ -295,15 +299,25 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
...
-	case "PRIVATE KEY":
-		return FileBackend, nil
-	case "TSS2 PRIVATE KEY":
-		return TPMBackend, nil
-	default:
-		return "", fmt.Errorf("unknown file type: %s", block.Type)
+	if json.Valid(b) {
+		var yubiData YubikeyData
+		err := json.Unmarshal(b, &yubiData)
+		if err != nil {
```
Foxboron, line 305:
```suggestion
		if err := json.Unmarshal(b, &yubiData); err != nil {
```

backend/yubikey.go
```diff
@@ -0,0 +1,319 @@
...
+			}
+			conf.PubKeyInfo = &newKeyInfo
+			h := md5.New()
+			h.Write(x509.MarshalPKCS1PublicKey(newKeyInfo.PublicKey.(*rsa.PublicKey)))
+			logging.Println(fmt.Sprintf("Created RSA2048 key MD5: %x", h.Sum(nil)))
+		}
+	} else {
+		// Key already setup for sbctl
+		pub = conf.PubKeyInfo.PublicKey
+	}
```
Foxboron, line 149: Invert the `if` and have the smaller case at the top.

backend/yubikey.go
```diff
@@ -0,0 +1,319 @@
...
+		NotBefore:          time.Now(),
+		NotAfter:           time.Now().AddDate(5, 0, 0),
+		Subject: pkix.Name{
+			Country:    []string{desc},
+			CommonName: desc,
+		},
+	}
+
+	h := md5.New()
+	h.Write(x509.MarshalPKCS1PublicKey(conf.PubKeyInfo.PublicKey.(*rsa.PublicKey)))
```
Foxboron, line 176: What is the intention of this?

config/config.go
```diff
@@ -125,20 +128,55 @@ func HasConfigurationFile(fs afero.Fs, file string) bool {
...
+
+	if err := yaml.Unmarshal(b, newConf); err != nil {
+		return nil, err
+	}
+
+	// TODO
+	// Should sbctl make the directories if they do not exist?
+	if err := VerifyConfig(fs, newConf); err != nil {
+		return nil, fmt.Errorf("invalid sbctl config; %w", err)
+	}
```
Foxboron, line 163: This doesn't belong in this PR. So lets remove it and just focus on the yubikey part.

config/config.go
```diff
@@ -125,20 +128,55 @@ func HasConfigurationFile(fs afero.Fs, file string) bool {
...
+	return nil
+}
+
+func NewConfig(fs afero.Fs, b []byte) (*Config, error) {
+	conf := &Config{}
 	if err := yaml.Unmarshal(b, conf); err != nil {
 		return nil, err
 	}
-	return conf, nil
+	var newConf = new(Config)
```
Foxboron, line 148: `conf := &Config{}` defines the config, so whats the intention here?

config/config.go
```diff
@@ -125,20 +128,55 @@ func HasConfigurationFile(fs afero.Fs, file string) bool {
...
+	if conf.Datadir != "" {
+		newConf = MkConfig(conf.Datadir)
+	} else {
+		newConf = DefaultConfig()
+	}
```
Foxboron, line 153: I'm not sure if want this here? What problem are you trying to solve with this?

config/config.go
```diff
@@ -125,20 +128,55 @@ func HasConfigurationFile(fs afero.Fs, file string) bool {
...
+	if err := VerifyConfig(fs, newConf); err != nil {
+		return nil, fmt.Errorf("invalid sbctl config; %w", err)
+	}
+
+	return newConf, nil
+}
+
+type YubiConfig struct {
+	PubKeyInfo *piv.KeyInfo
+	Overwrite  bool
```
Foxboron, line 170: Is this going to hardcode all key operations (PK, KEK and db) to the same yubikey? Or what is this doing at the top-level?

lsm/lsm.go
```diff
@@ -25,22 +25,29 @@ func LandlockRulesFromConfig(conf *config.Config) {
...
 			"/etc/localtime",
+			"/proc/sys/kernel/cap_last_cap",
+			"/usr/bin/systemd-ask-password",
 		).IgnoreIfMissing(),
 		landlock.RWFiles(
 			conf.GUID,
 			conf.FilesDb,
 			conf.BundlesDb,
+			//TODO will this be different for others for yubikeys?
+			"/usr/lib/libpcsclite_real.so.1",
```
Foxboron, line 49: I suspect this will be a bit tricky to figure out. Making a mental note to figure it out.

prompt/prompt.go
```diff
@@ -0,0 +1,58 @@
+package prompt
```
Foxboron, line 1: I'm not sure if I think we should solve yubikey and the prompting of user input in the same PR. This is complicated enough.

Could we introduce a `SBCTL_YUBIKEY_PIN` env variable to read the pin from as a temporary measure? Or just assume the yubikey is not locked?

Foxboron (changes requested):

tomis007:
Awesome thank you for the review! 

I didn't forget, I lost my yubikey recently which has made testing more difficult. I will hopefully have another one next week. I'm planning on restructuring the commits tomorrow to get rid of the intermediate ones and will update with your suggestions

prompt/prompt.go
tomis007: I can definitely break that off into another PR and do a temporary measure for pinstorage, however I would like to get the prompt feature also incorporated to avoid needing to store the PIN anywhere. I do explicitly set the yubikey policy to require the pin/touch always
```
PINPolicy:   piv.PINPolicyAlways,
TouchPolicy: piv.TouchPolicyAlways,
```

 I think a separate change would work better to help keep the changes introduced per commit / pr down

backend/yubikey.go
tomis007: The intention is to display a md5sum fingerprint for the public key in use (similar to how ykman displays the fingerprint)

lsm/lsm.go
tomis007: Agree, I installed the dependency on my arch system and a debian one and so far the required library has matched.

config/config.go
tomis007: Because of go's unmarshalling behavior and the current yaml parsing, when I tried to write my own sbctl yaml config my yubikey settings for the keytypes would be overwritten. The changes in this file fixed the parsing logic to not overwrite and ignore some of the settings. I'll remove them and make them in a separate PR

config/config.go
tomis007: From a high level, I've currently implemented the "createkey" method for PK/KEK/DB with yubikey support to all use the Signature slot 9C in the yubikey piv (https://docs.yubico.com/yesdk/users-manual/application-piv/slots.html). 

The purpose of `YubiConfig` is to track the state of the public key for the Signature Slot key pair when performing key signing or creating operations. Once a public key is stored in `PubKeyInfo` that key will be requested when performing a signing operation on the yubikey. When `NewYubikeyKey` creates a new PK/KEK/DB key X509 cert, it signs it with the same RSA2048 key in slot 9C. So really for the `yubikey` type what is currently implemented is that all three PK/KEK/DB keys are signed with the same RSA2048 key in slot 9C on the yubikey

tomis007:
Closing and moving to https://github.com/Foxboron/sbctl/pull/418 with a cleaner commit history :)

PR #417, opened 2025-02-08
ci.yaml: fixup OVMF_* for archlinux

pheiduck:


Foxboron:
So I pushed a fix that uses the renamed files in the tests. I'm not sure if the void CI is going to fail, so better to apply this change there :)

https://github.com/Foxboron/sbctl/commit/4386be42db8271aff477faf6769f3441f141d9b6

Thanks!

PR #418, opened 2025-02-08
Initial yubikey backend keytype support

tomis007:
Same PR for initial yubikey piv support with the intermediate commits removed.

I also removed:
* changes to the config parsing (will address in another pr)
* prompting for yubikey pin (right now uses the default PIN)

Hopefully this helps for reviewing / testing. I think I also addressed all your comments from the previous review. Thank you for your help!

I do think the prompting for the PIN is important to avoid needing to store the pin anywhere, so once we get this PR done I can rework and submit that one separately

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -3,9 +3,11 @@ package backend
 import (
 	"crypto"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"github.com/foxboron/sbctl/logging"
 	"io"
 	"os"
 	"path/filepath"
@@ -81,8 +83,6 @@ func (k *KeyHierarchy) GetKeyBackend(e efivar.Efivar) KeyBackend {
 		return k.KEK
 	case efivar.Db:
 		return k.Db
-		// case efivar.Dbx:
-		// 	return k.Dbx
 	default:
 		panic("invalid key hierarchy")
 	}
@@ -200,6 +200,8 @@ func createKey(state *config.State, backend string, hier hierarchy.Hierarchy, de
 		return NewFileKey(hier, desc)
 	case "tpm":
 		return NewTPMKey(state.TPM, desc)
+	case "yubikey":
+		return NewYubikeyKey(state.YubikeySigKeys, hier, desc)
 	default:
 		return NewFileKey(hier, desc)
 	}
@@ -255,6 +257,8 @@ func readKey(state *config.State, keydir string, kc *config.KeyConfig, hier hier
 		return FileKeyFromBytes(keyb, pemb)
 	case TPMBackend:
 		return TPMKeyFromBytes(state.TPM, keyb, pemb)
+	case YubikeyBackend:
+		return YubikeyFromBytes(state.YubikeySigKeys, keyb, pemb)
 	default:
 		return nil, fmt.Errorf("unknown key")
 	}
@@ -295,15 +299,24 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
 }
 
 func GetBackendType(b []byte) (BackendType, error) {
-	block, _ := pem.Decode(b)
-	// TODO: Add TSS2 keys
-	switch block.Type {
-	case "PRIVATE KEY":
-		return FileBackend, nil
-	case "TSS2 PRIVATE KEY":
-		return TPMBackend, nil
-	default:
-		return "", fmt.Errorf("unknown file type: %s", block.Type)
+	if json.Valid(b) {
+		var yubiData YubikeyData
+		if err := json.Unmarshal(b, &yubiData); err != nil {
+			logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
+			return "", err
+		}
+		return YubikeyBackend, nil
+	} else {
+		block, _ := pem.Decode(b)
+		// TODO: Add TSS2 keys
+		switch block.Type {
+		case "PRIVATE KEY":
+			return FileBackend, nil
+		case "TSS2 PRIVATE KEY":
+			return TPMBackend, nil
+		default:
+			return "", fmt.Errorf("unknown file type: %s", block.Type)
+		}
 	}
 }
 
@@ -322,6 +335,8 @@ func InitBackendFromKeys(state *config.State, priv, pem []byte, hier hierarchy.H
 		return FileKeyFromBytes(priv, pem)
 	case "tpm":
 		return TPMKeyFromBytes(state.TPM, priv, pem)
+	case "yubikey":
+		return YubikeyFromBytes(state.YubikeySigKeys, priv, pem)
 	default:
 		return nil, fmt.Errorf("unknown key backend: %s", t)
 	}
--- a/backend/yubikey.go
+++ b/backend/yubikey.go
@@ -1,0 +1,279 @@
+package backend
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/md5"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"github.com/foxboron/sbctl/config"
+	"github.com/foxboron/sbctl/hierarchy"
+	"github.com/foxboron/sbctl/logging"
+	"math/big"
+	"strings"
+	"time"
+
+	"github.com/go-piv/piv-go/v2/piv"
+)
+
+type YubikeyData struct {
+	Info      piv.KeyInfo `json:"Info"`
+	Slot      string      `json:"Slot"`
+	PublicKey string      `json:"PublicKey"`
+}
+
+var YK *piv.YubiKey
+
+type Yubikey struct {
+	keytype    BackendType
+	cert       *x509.Certificate
+	pubKeyInfo *piv.KeyInfo
+}
+
+func PIVKeyString(algorithm piv.Algorithm) string {
+	switch algorithm {
+	case piv.AlgorithmRSA2048:
+		return "RSA2048"
+	case piv.AlgorithmRSA3072:
+		return "RSA3072"
+	case piv.AlgorithmRSA4096:
+		return "RSA4096"
+	case piv.AlgorithmEC256:
+		return "EC256"
+	case piv.AlgorithmEC384:
+		return "EC384"
+	case piv.AlgorithmEd25519:
+		return "Ed25519"
+	case piv.AlgorithmX25519:
+		return "X25519"
+	default:
+		logging.Errorf("Unsupported Yubikey algorithm: %v", algorithm)
+		return ""
+	}
+}
+
+func NewYubikeyKey(conf *config.YubiConfig, hier hierarchy.Hierarchy, desc string) (*Yubikey, error) {
+	var pub crypto.PublicKey
+	if conf.PubKeyInfo != nil {
+		// Key already setup for sbctl
+		pub = conf.PubKeyInfo.PublicKey
+	} else {
+		// Find a YubiKey and open the reader.
+		if YK == nil {
+			var err error
+			YK, err = connectToYubikey()
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		keyInfo, err := YK.KeyInfo(piv.SlotSignature)
+		if err != nil {
+			return nil, err
+		}
+		// if there is a key and overwrite is set, create a new one
+
+		if keyInfo.PublicKey != nil {
+			// if there is a key and it is RSA2048 and overwrite is false, use it
+			if keyInfo.Algorithm == piv.AlgorithmRSA2048 && !conf.Overwrite {
+				logging.Println("RSA2048 Key exists in Yubikey PIV Signature Slot, using the existing key")
+				pub = keyInfo.PublicKey
+				conf.PubKeyInfo = &keyInfo
+			} else if !conf.Overwrite {
+				return nil, fmt.Errorf("yubikey key creation failed; %s key present in signature slot", PIVKeyString(keyInfo.Algorithm))
+			}
+		}
+		if keyInfo.PublicKey == nil || conf.Overwrite {
+			if conf.Overwrite {
+				logging.Warn("Overwriting existing key %s in Signature slot", PIVKeyString(keyInfo.Algorithm))
+			}
+
+			// Generate a private key on the YubiKey.
+			key := piv.Key{
+				Algorithm:   piv.AlgorithmRSA2048,
+				PINPolicy:   piv.PINPolicyAlways,
+				TouchPolicy: piv.TouchPolicyAlways,
+			}
+			logging.Println("Creating RSA2048 key...")
+			pub, err = YK.GenerateKey(piv.DefaultManagementKey, piv.SlotSignature, key)
+			if err != nil {
+				return nil, err
+			}
+			newKeyInfo, err := YK.KeyInfo(piv.SlotSignature)
+			if err != nil {
+				return nil, err
+			}
+			conf.PubKeyInfo = &newKeyInfo
+			h := md5.New()
+			h.Write(x509.MarshalPKCS1PublicKey(newKeyInfo.PublicKey.(*rsa.PublicKey)))
+			logging.Println(fmt.Sprintf("Created RSA2048 key MD5: %x", h.Sum(nil)))
+		}
+	}
+
+	auth := piv.KeyAuth{PIN: piv.DefaultPIN}
+	priv, err := YK.PrivateKey(piv.SlotSignature, pub, auth)
+	if err != nil {
+		return nil, err
+	}
+
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, _ := rand.Int(rand.Reader, serialNumberLimit)
+	c := x509.Certificate{
+		SerialNumber:       serialNumber,
+		PublicKeyAlgorithm: x509.RSA,
+		SignatureAlgorithm: x509.SHA256WithRSA,
+		NotBefore:          time.Now(),
+		NotAfter:           time.Now().AddDate(5, 0, 0),
+		Subject: pkix.Name{
+			Country:    []string{desc},
+			CommonName: desc,
+		},
+	}
+
+	h := md5.New()
+	h.Write(x509.MarshalPKCS1PublicKey(conf.PubKeyInfo.PublicKey.(*rsa.PublicKey)))
+	logging.Println(fmt.Sprintf("Creating %s key...\nPlease press Yubikey to confirm presence for key %s MD5: %x",
+		hier.String(),
+		PIVKeyString(conf.PubKeyInfo.Algorithm),
+		h.Sum(nil)))
+	derBytes, err := x509.CreateCertificate(rand.Reader, &c, &c, pub, priv)
+	if err != nil {
+		return nil, err
+	}
+
+	cert, err := x509.ParseCertificate(derBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Yubikey{
+		keytype:    YubikeyBackend,
+		cert:       cert,
+		pubKeyInfo: conf.PubKeyInfo,
+	}, nil
+}
+
+func YubikeyFromBytes(c *config.YubiConfig, keyb, pemb []byte) (*Yubikey, error) {
+	var yubiData YubikeyData
+	err := json.Unmarshal(keyb, &yubiData)
+	if err != nil {
+		logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
+		return nil, err
+	}
+
+	pubKeyB64, err := base64.StdEncoding.DecodeString(yubiData.PublicKey)
+	if err != nil {
+		return nil, err
+	}
+	pubKey, err := x509.ParsePKCS1PublicKey(pubKeyB64)
+	if err != nil {
+		logging.Errorf("Error parsing public key: %v\n", err)
+		return nil, err
+	}
+	yubiData.Info.PublicKey = pubKey
+
+	block, _ := pem.Decode(pemb)
+	if block == nil {
+		return nil, fmt.Errorf("no pem block")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse cert: %w", err)
+	}
+
+	c.PubKeyInfo = &yubiData.Info
+	return &Yubikey{
+		keytype:    YubikeyBackend,
+		cert:       cert,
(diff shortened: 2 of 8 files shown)


Foxboron:
Please stop making new PRs :) Rebase and force-push the branch in-place is much easier instead of leaving 3 PRs around in the project.

tomis007:
Sure, I'll do that going forward, I didn't know that worked in a github PR my bad!

tomis007:
Would it be helpful if rebased and force pushed this commit PR over https://github.com/Foxboron/sbctl/pull/413? I can then use this one for the prompt support PR

Foxboron:
> Would it be helpful if rebased and force pushed this commit PR over https://github.com/Foxboron/sbctl/pull/413? I can then use this one for the prompt support PR

Nah, no we can work in this branch :)

backend/backend.go
```diff
@@ -295,15 +299,24 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
...
+		// TODO: Add TSS2 keys
+		switch block.Type {
+		case "PRIVATE KEY":
+			return FileBackend, nil
+		case "TSS2 PRIVATE KEY":
+			return TPMBackend, nil
+		default:
+			return "", fmt.Errorf("unknown file type: %s", block.Type)
+		}
 	}
```
Foxboron, line 320: We don't need the else block, as the `if` block will return.

backend/yubikey.go
```diff
@@ -0,0 +1,279 @@
...
+		NotBefore:          time.Now(),
+		NotAfter:           time.Now().AddDate(5, 0, 0),
+		Subject: pkix.Name{
+			Country:    []string{desc},
+			CommonName: desc,
+		},
+	}
+
+	h := md5.New()
+	h.Write(x509.MarshalPKCS1PublicKey(conf.PubKeyInfo.PublicKey.(*rsa.PublicKey)))
```
Foxboron, line 140: This code is repeated, so lets make this into a function.

cmd/sbctl/create-keys.go
```diff
@@ -100,6 +107,7 @@ func RunCreateKeys(state *config.State) error {
 
 func createKeysCmdFlags(cmd *cobra.Command) {
 	f := cmd.Flags()
+	f.BoolVar(&OverwriteYubikey, "overwrite", false, "overwrite existing key if it exists in the Yubikey Signature slot")
```
Foxboron, line 110: Lets namespace this under `--yk-*` so `--yk-overwrite`.

cmd/sbctl/create-keys.go
```diff
@@ -33,7 +36,7 @@ var createKeysCmd = &cobra.Command{
 func RunCreateKeys(state *config.State) error {
 	if state.Config.Landlock {
 		lsm.RestrictAdditionalPaths(
-			landlock.RWDirs(filepath.Dir(filepath.Dir(filepath.Clean(state.Config.Keydir)))),
+			landlock.RWDirs(filepath.Dir(filepath.Clean(state.Config.Keydir))),
```
Foxboron, line 39: Mmm, why is this being changed?

backend/yubikey.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/go-piv/piv-go/v2/piv"
+)
+
+type YubikeyData struct {
+	Info      piv.KeyInfo `json:"Info"`
+	Slot      string      `json:"Slot"`
+	PublicKey string      `json:"PublicKey"`
+}
+
+var YK *piv.YubiKey
```
Foxboron, line 31: Any reason we can't throw the piv connetion into the `Yubikey` struct?

Foxboron (changes requested):
Thanks for splitting stuff up :) Makes easier to review this!

cmd/sbctl/create-keys.go
tomis007: I changed it trying to debug landlock with the yubikey, the change isn't actually necessary I reverted it back to what it was originally

backend/yubikey.go
tomis007: I tried to do that originally but I ended up needing to open up a piv connection and track piv state in `YubikeyFromBytes` during the signing workflow and it became more complicated. I ended up going with  the most simple approach to start

tomis007:
Updated with the requested changes, I also verified that using sbctl to create a key with the yubikey and signing files is working (for me at least) with the default pin

backend/yubikey.go
Foxboron: But can't we just have a pointer in the struct and use `connectToYubikey()` as a struct method when the actual card is needed? 

Everything seems fairly self-contained at the moment?

backend/yubikey.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	return yk, nil
+}
+
+func (f *Yubikey) Type() BackendType              { return f.keytype }
+func (f *Yubikey) Certificate() *x509.Certificate { return f.cert }
+
+func (f *Yubikey) Signer() crypto.Signer {
+	auth := piv.KeyAuth{PIN: piv.DefaultPIN}
```
Foxboron, line 225:
```suggestion
	auth := piv.KeyAuth{PIN: piv.DefaultPIN}
	if pin, found := os.LookupEnv("SBCTL_YUBIKEY_PIN"); found {
		auth = piv.KeyAuth{PIN: pin}
	}
```

Something like this, so you can escape the default pin.

backend/yubikey.go
```diff
@@ -0,0 +1,282 @@
...
+//	{
+//	   "Info": piv.pubKeyInfo,
+//	   "Slot": "signature",
+//	   "PublicKey": publicKey
+//	}
+func (f *Yubikey) PrivateKeyBytes() []byte {
+	yubiData := YubikeyData{
+		Info:      *f.pubKeyInfo,
+		Slot:      "signature",
+		PublicKey: base64.StdEncoding.EncodeToString(x509.MarshalPKCS1PublicKey(f.pubKeyInfo.PublicKey.(*rsa.PublicKey))),
```
Foxboron, line 259: We same `f.pubKeyInfo` here, why do we have `YubiConfig.PubKeyInfo` as well stored as part of the state?

Is this to prevent us from reading the information from the yubikey multiple times?

tomis007: It's also stored in the config to make sure when you want to overwrite or create a new one, it only creates a new one on signature slot in the yubikey once

backend/yubikey.go
tomis007: That approach works for creating keys, but the tricky part is in the Signer interface. the piv-go library only lets you have one handle to the yubikey open at once (https://github.com/go-piv/piv-go/blob/0383b0aa884b2b642e9e3446ea01ba22ccadc83a/v2/piv/piv.go#L107). When you  `defer yk-piv.Close()` after creating the handle in `Signer()` the signing operation happens after the function is returned and by then the handle is closed. If you leave the handle open you can't open a new handle to the yubikey for future operations.

I can make it work but I would likely need to change the Signer interface to return a function that gets called after signing operations to appropriately Close the handle. What do you think about that approach?

Foxboron:
I saw down to try this change and I can't seem to get it to work. The yubikey blinks for a press, but nothing happens when I press. I suspect the issue is on my side, but I found this a bit weird. Have to spend a little bit of time to figure out why this happens and the potential failure case.

tomis007:
Interesting, I have not encountered that behavior before with my yubikey.

Are you able to create any key on your yubikey in the piv slot with `ykman`? 
```bash
ykman piv info
# you can set the touch/pin policy too
ykman piv keys generate --algorithm RSA2048 9a pubkey.test
```
What happens if you update yubikey.go and change the PIN/Touch policy to never?:
```go
piv.Key{
    Algorithm:   piv.AlgorithmRSA2048,
    PINPolicy:   piv.PINPolicyNever,
    TouchPolicy: piv.TouchPolicyNever,
}
```

tomis007:
I just tested after resetting my yubikey `piv` data and it worked as expected:
```
$ ykman info
Device type: YubiKey 5C
Serial number: 12341234
Firmware version: 5.7.1
Form factor: Keychain (USB-C)
Enabled USB interfaces: OTP, FIDO, CCID

Applications
Yubico OTP      Enabled
FIDO U2F        Enabled
FIDO2           Enabled
OATH            Enabled
PIV             Enabled
OpenPGP         Enabled
YubiHSM Auth    Enabled

$ ykman piv reset
WARNING! This will delete all stored PIV data and restore factory settings. Proceed? [y/N]: y
Resetting PIV data...
Reset complete. All PIV data has been cleared from the YubiKey.
Your YubiKey now has the default PIN, PUK and Management Key:
        PIN:    123456
        PUK:    12345678
        Management Key: 010203040506070801020304050607080102030405060708


$ ./sbctl create-keys --keytype yubikey
Created Owner UUID 865d27c1-27a4-4b1d-bc5b-36ebf1d39124
Creating RSA2048 key...
Please press Yubikey to confirm presense
Created RSA2048 key MD5: 9a50e63ad7c616fb6aa8f0f5b6ebb680
Creating PK key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 9a50e63ad7c616fb6aa8f0f5b6ebb680
Creating KEK key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 9a50e63ad7c616fb6aa8f0f5b6ebb680
Creating db key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 9a50e63ad7c616fb6aa8f0f5b6ebb680
✓
Secure boot keys created!
```

How far did it get when you tested? I wonder if there is a missing dependency, do you have `pcsclite` installed?

tomis007:
Any luck testing? If you were unable to get it working I can take a look on a few different computers I have and try to reproduce the issue.

Foxboron:
Sorry!

I haven't managed to look at this again but I'll take a stab at this within the next couple of days :) Thanks!

Foxboron:
Hmmmmm, ran the same commands as you and reset the yubikey.

> couldn't initialize secure boot: command failed: smart card error 6d00

I suspect there is something wrong with scdaemon when it comes to exclusive vs shared access? I'll try and figure this out on my end

tomis007:
Hmm, I'll do some trouble shooting as well. What yubikey / firmware version are you using?

Foxboron:
```
Device type: YubiKey 5 NFC
Serial number: 9027747
Firmware version: 5.1.1
Form factor: Keychain (USB-A)
Enabled USB interfaces: OTP, FIDO, CCID
NFC transport is enabled

Applications	USB          	NFC
Yubico OTP  	Enabled      	Enabled
FIDO U2F    	Enabled      	Enabled
FIDO2       	Enabled      	Enabled
OATH        	Enabled      	Enabled
PIV         	Enabled      	Disabled
OpenPGP     	Enabled      	Enabled
YubiHSM Auth	Not available	Not available
```

The one I tried with

tomis007:
It seems like Yubico does not support upgrading/downgrading yubikey firmware, I haven't been able to trace down or reproduce your issue yet. Does your yubikey work as expected with `ykman`?

```
ykman piv info
ykman piv keys generate --algorithm RSA2048 9a pubkey.test
```

Foxboron:
I'll try and pick this up again. I do really want yubikey support solved for `sbctl` but I suspect I need to set off a bit more time to figure out why things are not working as expected.

tomis007:
I tried with a different Yubikey (5 USB A) to make sure everything was working on the branch still:
```
$ git clone git@github.com:tomis007/sbctl.git -b initial-yubikey-support
$ make
a2x --no-xmllint --asciidoc-opts="-f docs/asciidoc.conf" -d manpage -f manpage -D docs docs/sbctl.8.txt
a2x --no-xmllint --asciidoc-opts="-f docs/asciidoc.conf" -d manpage -f manpage -D docs docs/sbctl.conf.5.txt
fatal: No names found, cannot describe anything.
go build -ldflags="-X github.com/foxboron/sbctl.Version=" -o sbctl ./cmd/sbctl
$ sudo ./sbctl create-keys --keytype yubikey
Created Owner UUID 018ba2df-8734-4889-bcb5-43f3a52fc21f
Creating RSA2048 key...
Please press Yubikey to confirm presense
Created RSA2048 key MD5: 99bec93ddf3d20fcc208583b611700b2
Creating PK key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 99bec93ddf3d20fcc208583b611700b2
Creating KEK key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 99bec93ddf3d20fcc208583b611700b2
Creating db key...
Please press Yubikey to confirm presence for key RSA2048 MD5: 99bec93ddf3d20fcc208583b611700b2
✓
Secure boot keys created!
$ ykman info
Device type: YubiKey 5 NFC
Serial number: 52734083
Firmware version: 5.7.4
Form factor: Keychain (USB-A)
Enabled USB interfaces: OTP, FIDO, CCID
NFC transport is enabled

Applications    USB     NFC
Yubico OTP      Enabled Enabled
FIDO U2F        Enabled Enabled
FIDO2           Enabled Enabled
OATH            Enabled Enabled
PIV             Enabled Enabled
OpenPGP         Enabled Enabled
YubiHSM Auth    Enabled Enabled
```

This was the same error that you described but your Yubikey seems newer than the ones in the issue: https://github.com/go-piv/piv-go/issues/55. I haven't been able to reproduce the issue unfortunately and I have no older yubikeys to test against

Foxboron:
Judging by https://github.com/FiloSottile/yubikey-agent/tree/main I'm wondering if we should omit `KeyInfo` and just save the public key and make assumptions about key policies and so on. Would that work?

We could alternatively store it if the yubikey is of a recent version.

tomis007:
Oh nice find on the cause of the issue. I think it'd be ideal to ensure it works on older firmware versions too rather than just supporting newer ones. I'll take a look at how yubikey-agent manages older keys. It might take me a few days to get around to updating the pr

tomis007:
@Foxboron I updated and refactored to remove the KeyInfo calls so it should hopefully work with your yubikey now. I did some more refactoring to improve the code and the yubikey connection handling. It is ready for a review/testing

backend/backend.go
```diff
@@ -295,6 +300,14 @@ func GetKeyHierarchy(vfs afero.Fs, state *config.State) (*KeyHierarchy, error) {
 }
 
 func GetBackendType(b []byte) (BackendType, error) {
+	if json.Valid(b) {
+		var yubiData YubikeyData
+		if err := json.Unmarshal(b, &yubiData); err != nil {
```
Foxboron, line 305: You can probably just omit the variable assignment if the intention is to just check if the format serializes into `YubikeyData`.

> if err := json.Unmarshal(b, &YbuikeyData{}); err != nil {

backend/yubikey.go
```diff
@@ -0,0 +1,218 @@
...
+	algorithm     piv.Algorithm
+	pinPolicy     piv.PINPolicy
+	touchPolicy   piv.TouchPolicy
+}
+
+func NewYubikeyKey(yubikeyReader *config.YubikeyReader, hier hierarchy.Hierarchy) (*Yubikey, error) {
+	cert, err := yubikeyReader.GetPIVKeyCert()
+	if err != nil {
+		if !errors.Is(err, piv.ErrNotFound) {
+			return nil, err
```
Foxboron, line 48: Lets wrap this error!
```suggestion
			return nil, fmt.Errorf("failed finding yubikey: %v", err)
```

backend/yubikey.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	if cert != nil {
+		// if there is a key and it is RSA2048 and overwrite is false, use it
+		switch yubiPub := cert.PublicKey.(type) {
+		case *rsa.PublicKey:
+			// RSA2048 Public Key
+			if yubiPub.N.BitLen() == 2048 && !yubikeyReader.Overwrite {
```
Foxboron, line 57: Mm, we hardcode `2048` RSA keys, but `sbctl` explicitly hardcodes `4096`.

Should we go for `4096` instead? I believe all yubikeys from 4 and onwards support this just fine.

backend/yubikey.go
```diff
@@ -0,0 +1,218 @@
...
+		touchPolicy:   piv.TouchPolicyAlways,
+	}, nil
+}
+
+func YubikeyFromBytes(yubikeyReader *config.YubikeyReader, keyb, pemb []byte) (*Yubikey, error) {
+	var yubiData YubikeyData
+	err := json.Unmarshal(keyb, &yubiData)
+	if err != nil {
+		logging.Errorf("Error unmarshalling Yubikey: %v\n", err)
+		return nil, err
```
Foxboron, line 145: Return this as an error instead of `logging.Errorf` I think.

config/config.go
```diff
@@ -133,12 +139,106 @@ func NewConfig(b []byte) (*Config, error) {
 	return conf, nil
 }
 
+// A type to wrap piv.Yubikey to manage the yubikey handle
+type YubikeyReader struct {
+	key       *piv.YubiKey
+	Overwrite bool
+}
```
Foxboron, line 146: Lets throw this into `config/yubikeyreader.go` instead of `config/config.go`

config/config.go
```diff
@@ -133,12 +139,106 @@ func NewConfig(b []byte) (*Config, error) {
...
+			return nil, err
+		}
+	}
+	return y.key.PrivateKey(slot, public, auth)
+}
+
+func (y *YubikeyReader) connectToYubikey() error {
+	if y.key != nil {
+		return nil
+	}
```
Foxboron, line 178: As we are just returning `nil` here, we can remove all the `if y.key == nil` checks earlier in the code.

lsm/lsm.go
```diff
@@ -41,6 +41,8 @@ func LandlockRulesFromConfig(conf *config.Config) {
 			conf.GUID,
 			conf.FilesDb,
 			conf.BundlesDb,
+			//TODO will this be different for others for yubikeys?
+			"/usr/lib/libpcsclite_real.so.1",
```
Foxboron, line 45: I think we'll just commit this as-is and we'll figure out where it doesn't work in the future when we get bug reports :)

Foxboron:
Generally I think this is shaping up nicely. I'm a bit concerned as there is not a good way to test this at all. So I might merge this and see if I can find some time to check if I can do a small test setup before release.

Else I'll be a little bit dependent on you being willing to help debug issues :)

backend/yubikey.go
lespea: Should use `%w` not `%v` when wrapping errors

Foxboron: @lespea only if you want the caller to depend on the error "type". And in this case we don't know yet.

lsm/lsm.go
tomis007: Agreed

config/config.go
```diff
@@ -133,12 +139,106 @@ func NewConfig(b []byte) (*Config, error) {
...
+					}
+				}
+			}
+		}()
+		select {
+		case <-time.After(90 * time.Second):
+			return fmt.Errorf("timeout waiting for yubikey")
+		case cards = <-c:
+		}
+	}
```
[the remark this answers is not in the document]
tomis007: I think this pattern is fairly common in go (https://gobyexample.com/timeouts / https://go.dev/blog/concurrency-timeouts). I did refactor it to ensure that the goroutine returns when the timeout finishes (before the goroutine kept polling while the program exited)

I don't think the 
```
for {
    select {
    ...
    }
}
```
structure works because the select timeout is reset after each 500 Millisecond and it doesn't end up timing out. I think I might just be misunderstanding though how you want to refactor it. I think you need a goroutine to do the polling with the timeout.

tomis007:
I added all the requested changes, except the polling on the yubikey. I am happy to help with future issues! Let me know what you think on the updated code

Were you able to create / sign a uki with your yubikey?

One thing that I am tracking is that I think if the pin is non default there may be an issue with the hook after a uki rebuild gets triggered. I think that's a future issue to figure out though as this pr as already quite large

Foxboron:
>  I think that's a future issue to figure out though as this pr is already quite large and eventually I'd like to be able to prompt for the pin instead of storing it

Yes. If we just keep this to the basic yubikey functionality and then try to expend with pin reading later it will be much simpler for us :)

The entire process of prompting for information when we are trying to run interactively is probably going to be tricky. Which was my primary motivation for splitting this up.

config/config.go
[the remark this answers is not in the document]
tomis007: Ah yeah that makes sense if the goal is to get rid of the goroutine. I updated it to poll without the go call()

tomis007:
Requested changes are addressed and I confirmed yubikey creation/signing is working for my two yubikeys (both are on 5.7.X)

tomis007:
Sure, I'll clean it up in the next day or two, thanks for all the help!

tomis007:
I cleaned up the commit history. I thought about breaking it into a few different commits, but the yubikey support did not have a clean breakdown and the commits in between were not really functional and in a buggy/incomplete state. I ended up just rebaseing to the current state and fixuped the yubikey specific changes.

Foxboron:
I think that is completely fair :)

I'll do a last review and test, hopefully tonight.

README.md
```diff
@@ -261,3 +261,32 @@ Enrolled bundles:
 Generating EFI bundles....
 Wrote EFI bundle /efi/EFI/Linux/linux-linux.efi
 ```
+
+
+## Yubikey Support
+To create a signing key with a yubikey, use the `--keytype yubikey` flag with the `create-key` command. For example:
+
```
Foxboron, line 268: Much appreciated, this also needs to go into the man page, but I can do that for you if you want?

tomis007: Sure that'd be great if you want to add to the man page. I did notice in the man page a few other options are missing too: like for create keys a few options aren't  specified:
```
**create-keys**::
        Creates a set of signing keys used to sign EFI binaries. Currently, it
        will create the following keys:
        * Platform Key
        * Key Exchange Key
        * Signature Database Key

        *-e*, *--export*;;
                The directory to persist the exported keys.

        *-d*, *--database-path*;;
                Path to save the GUID file when generating keys.
```
vs
```
sbctl create-keys -h
Create a set of secure boot signing keys

Usage:
  sbctl create-keys [flags]

Flags:
  -d, --database-path string   location to create GUID file
      --db-keytype string      db key type (default: file)
  -e, --export string          export file path
  -h, --help                   help for create-keys
      --kek-keytype string     KEK key type (default: file)
      --keytype string         key type for all keys
      --pk-keytype string      PK key type (default: file)
      --yk-overwrite           overwrite existing key if it exists in the Yubikey Signature slot

Global Flags:
      --config string      Path to configuration file
      --debug              Enable verbose debug logging
      --disable-landlock   Disable landlock sandboxing
      --json               Output as json
      --quiet              Mute info from logging
```

Foxboron: I'll solve that then :) thanks!

Foxboron (approved):

Foxboron:
The CI does not like this. But I'll try and figure it out. It builds fine locally

PR #419, opened 2025-02-16
various fixes

pheiduck:
to match on:
https://github.com/Foxboron/sbctl/blob/master/config/config.go
https://github.com/Foxboron/go-uefi/blob/master/efi/util/guid.go


Foxboron:
Thanks!

PR #473, opened 2025-10-31
Filter non yubikeys

scj643:
Filter out non yubikey smart card readers.

I have a Thinkpad T470s that has a built in smartcard reader and this patch allows it to work by ignoring any readers that aren't a yubikey.


Foxboron:
Thanks!

Issue #462, opened 2025-10-06
Cannot sign kernel with sbctl after kernel reinstallation - failed validating signature: crypto/rsa: verification error


zhihuiyuze:
Hi everyone,

I'm experiencing a persistent issue with sbctl after a failed kernel update, and I need help understanding what went wrong and how to recover my TPM auto-unlock setup.

During a pacman kernel update, the process hung and I forcefully terminated it. After reinstalling the kernel, the automated signing workflow consistently fails with:

`failed validating signature: crypto/rsa: verification error `
The log at that time was the same as it is now by using mkinitcpio
```

[root@archlinux-laptop ~]# mkinitcpio -P
==> Building image from preset: /etc/mkinitcpio.d/linux-g14.preset: 'default'
==> Using default configuration file: '/etc/mkinitcpio.conf'
  -> -k /boot/vmlinuz-linux-g14 -U /boot/EFI/Linux/arch-linux-g14.efi
==> Starting build: '6.16.8-arch3-1.1-g14'
  -> Running build hook: [base]
  -> Running build hook: [systemd]
  -> Running build hook: [autodetect]
  -> Running build hook: [modconf]
  -> Running build hook: [block]
  -> Running build hook: [sd-encrypt]
==> WARNING: Possibly missing firmware for module: 'qat_6xxx'
  -> Running build hook: [lvm2]
  -> Running build hook: [filesystems]
  -> Running build hook: [keyboard]
  -> Running build hook: [fsck]
  -> Running build hook: [sd-vconsole]
==> Generating module dependencies
==> Creating zstd-compressed initcpio image
  -> Early uncompressed CPIO image generation successful
==> Initcpio image generation successful
==> Creating unified kernel image: '/boot/EFI/Linux/arch-linux-g14.efi'
  -> Using ukify to build UKI
  -> Using cmdline file: '/etc/kernel/cmdline'
Using config file: /etc/kernel/uki.conf
+ /usr/lib/systemd/systemd-sbsign sign --private-key /var/lib/sbctl/keys/db/db.key --certificate /var/lib/sbctl/keys/db/db.pem /boot/vmlinuz-linux-g14 --output /tmp/linux-signedbxju6xsr
Wrote signed PE binary to /tmp/linux-signedbxju6xsr
[...]

PR #480, opened 2025-11-25
Fix parsing for EFI binaries signed by `systemd-sbsign`

louiswpf:
`sbctl sign` failed verification when encountering an EFI binary signed
by `systemd-sbsign`, yielding the error `failed validating signature:
crypto/rsa: verification error`.

The cause was `systemd-sbsign` adding the S/MIME Capabilities attribute
to the signature [1].

The issue is resolved by updating go-uefi. The fix was incidental: the
upstream change (Foxboron/go-uefi@b82008d) was intended for Authenticode
timestamp support. This required implementing an explicit ASN.1 SET
wrapper for authenticated attributes, which coincidentally fixed the
S/MIME Capabilities parsing error.

This issue was reported in #462, although the author ultimately closed
it believing the problem was a configuration error on their part.

[1]: https://github.com/systemd/systemd/blob/8c22cb264d6c29b974132c7e8366eb533ec378f0/src/sbsign/sbsign.c#L357

Fixes: #462


Foxboron:
Oh! 
This makes a lot of sense. Thanks!

PR #485, opened 2026-01-03
installkernel: remove bashisms

Proto1337:
Convert the installkernel script to POSIX sh by removing bash-specific syntax
that caused failures on systems where /bin/sh is not bash (e.g. dash):

- Replacing `[[…]]` with POSIX `[…]` syntax (SC3010)
- Using `=` instead of `==` for string comparisons (SC3014)
- Replacing `$EUID` with `$(id -u)` (SC3028)
- Replacing `echo -e` and "\e" escapes with `printf` and "\033" (SC3037)
- Quoting variables to prevent globbing and word splitting (SC2086)

After this change, the script runs correctly with /bin/sh symlinked to dash.  
It also passes ShellCheck with no POSIX warnings.

### DIFF
--- a/contrib/installkernel/91-sbctl.install
+++ b/contrib/installkernel/91-sbctl.install
@@ -7,30 +7,32 @@ ver=${1}
 img=${2}
 
 die() {
-	echo -e " ${NOCOLOR-\e[1;31m*\e[0m }${*}" >&2
-	exit 1
+  printf ' \033[1;31m*\033[0m %s\n' "$*" >&2
+  exit 1
 }
 
 einfo() {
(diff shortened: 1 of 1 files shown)


Proto1337:
There isn't a single reason.
I saw that the kernel-install script was POSIX-compliant so I decided to do the same here.

Honestly changing the shebang to `#!/usr/bin/env bash` would already do the trick. The installkernel scripts in Gentoo also use `#!/usr/bin/env bash` as shebang.

Foxboron (approved):

Foxboron:
Thanks!

PR #492, opened 2026-02-10
fix(kernel-install): support grub layout

### LINKED ISSUE #491: `91-sbctl.install` fails with `layout=grub` -- tries to sign a non-existent BLS path
Might be related to my #490 setup, but 91-sbctl.install unconditionally constructs the image path as:
https://github.com/Foxboron/sbctl/blob/eba782214ed52257402ac05f6942368d9cf739d0/contrib/kernel-install/91-sbctl.install#L10

This assumes a BLS layout where `kernel-install` copies the kernel into `$BOOT/<machine-id>/<version>/linux`. There is a code path for `layout=uki`, but none for `layout=grub`.

when the layout is grub kernel-install doesn''t copy the kernel to the entry direwctory but installs it to `/boot/` (in my case `/boot/kernel-6.18.9-gentoo-dist`) and `grub-mkconfig` picks it up from there. The script then tries to sign a path that was never created and fails:
```bash
sbctl: Signing kernel /efi/70e379a7171d4167a22b44ac74864265/6.18.9-gentoo-dist/linux
/efi/70e379a7171d4167a22b44ac74864265/6.18.9-gentoo-dist/linux does not exist
/usr/lib/kernel/install.d/91-sbctl.install failed with exit status 1.
```

I think just adding (if i am not forgetting anything):
```bash
if [ "$KERNEL_INSTALL_LAYOUT" = "grub" ]; then
    IMAGE_FILE="$KERNEL_IMAGE"
```
will be a good enough of a fix tbh

Deftera186:
Hi, I have added an elif branch for `layout=grub` that uses `$KERNEL_IMAGE`, which points to the actual kernel image. Tested it on Gentoo and the kernel now emerges successfully.

Fixes #491

### DIFF
--- a/contrib/kernel-install/91-sbctl.install
+++ b/contrib/kernel-install/91-sbctl.install
@@ -4,10 +4,7 @@
 COMMAND="$1"
 KERNEL_VERSION="$2"
 ENTRY_DIR_ABS="$3"
-# shellcheck disable=SC2034  # Unused variables left for readability
 KERNEL_IMAGE="$4"
-
-IMAGE_FILE="$ENTRY_DIR_ABS/linux"
 
 if [ "$KERNEL_INSTALL_LAYOUT" = "uki" ]; then
 	UKI_DIR="$KERNEL_INSTALL_BOOT_ROOT/EFI/Linux"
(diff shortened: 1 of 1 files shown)


contrib/kernel-install/91-sbctl.install
```diff
@@ -23,6 +20,10 @@ if [ "$KERNEL_INSTALL_LAYOUT" = "uki" ]; then
 	else
 		IMAGE_FILE="$UKI_DIR/$KERNEL_INSTALL_ENTRY_TOKEN-$KERNEL_VERSION.efi"
 	fi
+elif [ "$KERNEL_INSTALL_LAYOUT" = "grub" ]; then
```
Foxboron: Please add a comment that references the gentoo part of this. I don't think anyone else is doing this.

Foxboron:
Thanks!

PR #503, opened 2026-07-16
fix: Restrict access to imported key material

Fahrj:
Without this change imported key-material is copied as-is, which could be extremely permissive. Since we know we are handling confidential key-material, let's restrict the access to the imported files.


Foxboron:
Thanks!

PR #504, opened 2026-07-16
fix(yubikey): Handle errors and multiple smartcards

Fahrj:
This commit generally improves the detection of YubiKeys, in detail:

-Errors when accessing smart cards with `piv.Cards()` are no longer suppressed
-Properly handle cases where there are multiple non-yubikey smartcards
-Code re-structure to avoid `goto` and make it more maintainable


Foxboron:
Thanks!

PR #505, opened 2026-07-29
Implement support for partial keys

### LINKED ISSUE #356: Image signature/verification fails if PK or KEK are missing
After updating to the latest version (0.15.4) and running `sbctl setup --migrate` as instructed, if the PK and/or KEK is not present, the `list-files`, `sign`, `sign-all` and `verify` operations fail with `/var/lib/sbctl/keys/KEK/KEK.key: no such file or directory` (and, in order, `KEK/KEK.pem`, `PK/PK.key` and ` PK/PK.pem` if they are also missing).

There is no reason for the KEK, let alone the PK, to be present on the system during regular operation, or to even be in possession of a local administrator. Signing an image only requires the db key, and verifying the signatures only the public part of it.

In fact, the files only need to exist and be valid keys (empty files lead to a SEGV in `backend.GetBackendType`), not necessarily be the actual enrolled PK and KEK (e.g. copying the db key over in all three locations works), for `sign` / `sign-all` to succeed and produce good signatures ; demonstrating it does not actually need these keys.

These operations should therefore not require the PK and KEK to be present in the filesystem.

Fahrj:
While this PR admittedly is rather massive, I tried to split it into digestible commits for easier review.

- The first three commits implement lazy loading of key backends from disk, namely PK, KEK, and db.
This will already close #356.

- The next three commits change the use of `KeyHierarchy` by turning a few functions into struct methods, which better fits the understanding I got from reading the code: after initializing a new object with `NewKeyHierarchy` it can be filled with `ReadKey`, `CreateKey`, or `ImportKey`, and then be used for operations such as sign/verify/save.
Please let me know if you agree with this approach.

- The next three commits prepare and implement partial key generation by adding `--partial` to `sbctl create-keys`
- The last commit simplifies CLI input checks, which I stumbled upon when implementing partial key support when creating keys

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -76,21 +76,21 @@ func (k *KeyHierarchy) GetKeyBackend(e efivar.Efivar) (KeyBackend, error) {
 	switch e {
 	case efivar.PK:
 		if k.pk == nil {
-			if k.pk, err = readKey(k.state, k.state.Config.Keydir, k.state.Config.Keys.PK, hierarchy.PK); err != nil {
+			if k.pk, err = k.ReadKey(hierarchy.PK); err != nil {
 				return nil, err
 			}
 		}
 		return k.pk, nil
 	case efivar.KEK:
 		if k.kek == nil {
-			if k.kek, err = readKey(k.state, k.state.Config.Keydir, k.state.Config.Keys.KEK, hierarchy.KEK); err != nil {
+			if k.kek, err = k.ReadKey(hierarchy.KEK); err != nil {
 				return nil, err
 			}
 		}
 		return k.kek, nil
 	case efivar.Db:
 		if k.db == nil {
-			if k.db, err = readKey(k.state, k.state.Config.Keydir, k.state.Config.Keys.Db, hierarchy.Db); err != nil {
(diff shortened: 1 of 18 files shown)


Fahrj:
Looking through the code I stumbled upon a small bug and added a fixup: e54b8f3bf01edd2a53ab176b1a40689661ff6547

Foxboron:
Quickly reviewing this code and I have no immediate comments. I'll need to sit down a bit more with this and test run things as there is a lack of test coverage in `sbctl` for the cli parts.

Fahrj:
> I'll need to sit down a bit more with this and test run things as there is a lack of test coverage in sbctl for the cli parts.

If I can help with something, let me know!


>Looking through the code I stumbled upon a small bug and added a fixup: https://github.com/Foxboron/sbctl/commit/e54b8f3bf01edd2a53ab176b1a40689661ff6547

The fixup unfortunately does not fit, because it uses a method which is introduced by a later commit. I added a proper fixup instead and renamed the initial commit to better reflect that it is trying to keep the functionality of `setup`, at least as far as I understand.

